PR #10, opened 2024-01-10
Assignment 1 - Cole Stewart

colestewart12:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,7 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return ((numberCorrectAnswers * 1.0) / (numberOfQuestions * 1.0) * 100) + "%";
```
Matthew-Powers, line 24: Should be able to only use a single * 1.0
```
        return ((numberCorrectAnswers * 1.0) / (numberOfQuestions) * 100) + "%";
```

Matthew-Powers (commented):
Slight code change suggestion

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
colestewart12: Thank you!

PR #12, opened 2024-01-11
Assignment 1 - Matthew Powers

Matthew-Powers:
Fixed the tests failing, and also fixed the program crashing


Assignment1/PrincessBrideTrivia/PrincessBrideTrivia.Tests/PrincessBrideTrivia.Tests.csproj
BenjaminMichaelis, line 1: Oops! looks like your project got moved around with it's folders. We expect the project (the .sln, then each project folder) to stay in the `PrincessBrideTrivia` folder like it did when given to you so that the automated tests can run against it properly.

Tvor369:
I will review this

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,15 +21,24 @@ public static void Main(string[] args)
...
+        return (1.0 * numberCorrectAnswers / numberOfQuestions * 100) + "%";
     }
 
     public static bool AskQuestion(Question question)
     {
         DisplayQuestion(question);
 
         string userGuess = GetGuessFromUser();
-        return DisplayResult(userGuess, question);
+        if(userGuess == "1" || userGuess == "2" || userGuess == "3")
```
Tvor369, line 32: You could use a while loop with your if statement here to repeat the question until they give an answer between 1 and 3. Have a bool variable that changes once a valid input is reached and breaks the while loop.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,15 +21,24 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return (1.0 * numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
Tvor369, line 24: I choose to cast both of the passed in variables as doubles and then cast the whole equation back to an int at the end.

Tvor369 (commented):
Assignment Details
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit (no partner)
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔
Extra Credit
Come up with a new feature for the application
Issue 3: Add a new feature request implemented ✔
Issue 3: Feature request unit tested ✔

Over good work. This may be my own user error with github but the code that I pulled down did not include the extra credit changes in the solution, but seeing it here it looks like it would work.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
Matthew-Powers: That seems like unnecessary extra code rather than just casting it once. Also, what if the user wants a more accurate percentage?

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
Matthew-Powers: That feels like I'm adding extra code for no reason

Matthew-Powers (commented):
Reviewed comments

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -71,6 +71,35 @@ public void GetFilePath_ReturnsFileThatExists()
...
+
+        //Assert
+        if(userChoice == "1")
+        {
+            Assert.AreEqual(result, "Valid input");
+        }
+        else
+        {
+            Assert.AreEqual(result, "Invalid input, please enter 1, 2, or 3");
+        }
```
BenjaminMichaelis, line 100: Remember, we want to avoid conditionals in our tests. This should be split into two separate tests

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
BenjaminMichaelis, line 32:
```suggestion
        if (int.TryParse(userGuess, out int guess) && 1 <= guess && guess <= 3)
```
Another way to accomplish this if you wanted to use int comparison instead of string comparisons

PR #25, opened 2024-01-15
Assignment1 - Trevor Stewart

Tvor369:


Tfelchlin13 (commented):
You can simplify this down to only casting one

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,8 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return (int)((numberCorrectAnswers / (double)numberOfQuestions) * 100) + "%";//Needed to use doubles
```
BenjaminMichaelis, line 24:
```suggestion
        return (double)((numberCorrectAnswers / numberOfQuestions) * 100) + "%";//Needed to use doubles
```
This can be simplified, and the `(int)` is unnecessary as the number is already an `int`

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,8 @@ public static void Main(string[] args)
...
+        //return (numberCorrectAnswers / numberOfQuestions * 100) + "%";//problem line
```
BenjaminMichaelis, line 25:
```suggestion
```
unnecessary code like this doesn't need to be committed for the final commit just to keep the code base a bit cleaner

PR #32, opened 2024-01-15
Assignment1 - Ethan Guerin

EthanAGi:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,7 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
MattA41, line 24: why did you use a double over a float?

EthanAGi (commented):
Good idea! I'll make that change!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
BenjaminMichaelis, line 1: This program when run has a runtime error

EthanAGi: I believe I fixed it!

PR #37, opened 2024-01-16
Assignment 1 - Paul Aguilar

pja6:
Assignment Completed
New Feature Suggestion: "Which character are you?" type output based on quiz results

### DIFF
--- a/ClassDemo/ClassDemo.Tests/UnitTest1.cs
+++ b/ClassDemo/ClassDemo.Tests/UnitTest1.cs
@@ -1,0 +1,12 @@
+dotnet namespace ClassDemo.Tests;
+
+public class UnitTest1
+{
+    [Fact]
+    public void Test1()
+    {
+        Program.Main(string[] args){
+            
+        };
+    }
+}
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -86,6 +86,8 @@ public class Program
             question.Answers[1] = answer2;
             question.Answers[2] = answer3;
             question.CorrectAnswerIndex = correctAnswerIndex;
+
+            questions[i] = question;
         }
         return questions;
     }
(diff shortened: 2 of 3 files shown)


connerhanson:
I'll review this one.

ClassDemo/ClassDemo.Tests/UnitTest1.cs
connerhanson, line 1: Looks like you pushed the class demo into the same branch as Assignment 1, though it doesn't effect the assignment I thought I should mention it.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -17,11 +17,27 @@ public static void Main(string[] args)
             }
         }
         Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+        Console.WriteLine("Your character match is: " + UserCharacterResult(numberCorrect,questions.Length) + "!");
     }
 
-    public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+    public static string GetPercentCorrect(double numberCorrectAnswers, double numberOfQuestions)
```
connerhanson, line 23: I never thought to change the method variable types, that's a good way to fix the problem.  One other way that's a little simpler would be to cast a double or float into the return statement, keeping the method as integers. Though both ways work well, I'll add a suggestion where this could be applied just in case you like that idea.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -17,11 +17,27 @@ public static void Main(string[] args)
...
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return Math.Round((numberCorrectAnswers / numberOfQuestions * 100),2) + "%";
```
connerhanson, line 25:
```suggestion
        return ((float)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
If used I would recommend reverting the method variable types to integers.

connerhanson (commented):
Good Job overall, it was nice seeing different ways students went about solving the problems in this assignment.  I liked the extra credit implementation as well!

    Pull Request title includes "Assignment 1" in the title ✔
    Partner has also made a commit ✔❌ (NA since they worked solo)
    Pull Request targets Assignment 1 branch ✔
    Issue 1: Application no longer crashes ✔
    Issue 2: Unit test properly passes ✔

    Issue 3: Add a new feature request implemented ✔
    Issue 3: Feature request unit tested ✔

PR #53, opened 2024-01-22
Assignment2 - Ethan Guerin

EthanAGi:


Purplepigs0129:
Reviewing

Logger/Logger.csproj
```diff
@@ -1,8 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>11.0</LangVersion>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
```
Purplepigs0129, line 5: I tried this in my own code and got an error on it surprised that you got it to work. Great Job!

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,31 @@
...
+using System.Globalization;
+
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
+    public static void Error(this BaseLogger? logger, string message, params object[] args)
+    {
+        ArgumentNullException.ThrowIfNull(logger, nameof(logger));
+        logger.Log(LogLevel.Error, string.Format(CultureInfo.InvariantCulture, message, args));
```
Purplepigs0129, line 11: Is this needed here? 
Mine is set to null but is this referencing the datetime format?

Logger/LogFactory.cs
```diff
@@ -1,10 +1,22 @@
...
+    }
+    public BaseLogger? CreateLogger(string className)
+    {
+        if(fileLoggerPath != null) {
+            var fileLogger = new FileLogger(fileLoggerPath)
+            {
+                ClassName = className
+            };
+            return fileLogger;
+        }
```
Purplepigs0129, line 19: would an else statement be better here because otherwise this method when the fileLoggerPath is not null will return the logger and null in one instance? could be wrong on this just wondering

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,22 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Logger.Tests;
 
 [TestClass]
 public class FileLoggerTests
 {
+    [TestMethod]
+    public void FileLogger_Log_MessageAppendedToFile()
```
Purplepigs0129, line 10: I think this test method is named incorrectly should be method under test(so Log) then condition(so MessageAppendsToFile) and then success or fail

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -8,15 +8,10 @@ namespace Logger.Tests;
...
     public void Error_WithNullLogger_ThrowsException()
     {
-        // Arrange
-
-        // Act
-        //BaseLoggerMixins.Error(null, "");
-
-        // Assert
+        BaseLogger? logger = null;
+        Assert.ThrowsException<ArgumentNullException>(() => logger.Error(""));
```
Purplepigs0129, line 14: Curious as to the addition of this, I just used the commented part and nothing else but I could see how this works.

Purplepigs0129 (commented):
The current date/time ✔
The log level ✔
The message ✔
Error, ✔
Warning, ✔
Information, and ✔
Ensure that you turn on EnforceCodeStyleInBuild ❌
Extra Credit
Could maybe add a few more unit tests but you are testing most of everything from what I can see and they are all passing, Great Work!

Logger.Tests/BaseLoggerMixinsTests.cs
EthanAGi: Good point, I made the change but I do need to make a TestLogger in each Test.

Logger.Tests/FileLoggerTests.cs
EthanAGi: Thank you! I left a placeholder in lol.

Logger/BaseLoggerMixins.cs
EthanAGi: This was a suggestion I was receiving after turn in. I looked into it, and it's formatting the string for other systems. I found it interesting and it fixed the error.

Logger/LogFactory.cs
EthanAGi: I made the change! I actually used ArgumentNullException.ThrowIfNull(nameof(fileLoggerPath)), because when implementing an if statement that triggered an error.

Logger/Logger.csproj
EthanAGi: Thank you! Honestly, it wasn't working until the last push, where I tried it and for some reason it worked. I did have to redownload .Net 8 something like 3 times but it eventually worked!

StrangeRanger:
@Purplepigs0129

> Ensure that you turn on EnforceCodeStyleInBuild ❌

When it comes to the analyzers/configuration-based requirements, make sure to provide a little description, such as which projects don't contain them.

Example:

> Ensure that you turn on EnforceCodeStyleInBuild ❌ => Not set in the test project

PR #54, opened 2024-01-23
Assignment2 - Joshua Lester

Joshua-Lester3:


timmytaters:
Starting review

timmytaters:
Assignment 2 Reviewing Joshua, by Tim
The current date/time ✔
The log level ✔
The message ✔
The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ❌ - ConfigureFileLogger is not used within CreateLogger
Error, ✔
Warning, ✔
Information, and ✔
Turn on Nullability (Nullable) ❌ - When I downloaded the code, nullability was off for logger
Extra Credit
Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses System.Console or System.Diagnostics.Trace ❌ - I did not see an additional logger
Implement the factory pattern using static abstract methods on a Logger interface instead. - ❌

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -26,12 +26,93 @@ public void Error_WithData_LogsMessage()
         var logger = new TestLogger();
 
         // Act
-        //logger.Error("Message {0}", 42);
+        logger.Error("Message 42");
```
timmytaters, line 29: I would recommend reverting this since it was given as such, so it is likely intended to work with the separate args

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,77 @@
...
+        Assert.IsTrue(containsExpectedOutput);
+    }
+
+    // DoesFileLoggerLog is a public, static method to reduce repeated code
+    // across FileLoggerTests.cs and LogFactoryTests.cs
+    public static bool DoesLoggerLog(BaseLogger? fileLogger, string filePath, string message)
+    {
+        // Act
+        fileLogger?.Log(LogLevel.Information, message);
+        string[] allLines = File.ReadAllLines(filePath);
```
timmytaters, line 69: While this does check the last line, maybe add a check that the rest of the doc/previous logs are preserved?

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,9 +1,62 @@
...
 
 [TestClass]
 public class LogFactoryTests
 {
-
+    [TestMethod]
+    public void CreateLogger_NonNullInputString_ReturnsBaseLogger()
+    {
+        // Arrange
+        LogFactory logFactory = new();
```
timmytaters, line 14: It seems like this constructor is called for many of these tests, possibly add a "[TestInitialize]" section so that it does it before every test?

Logger.Tests/Logger.Tests.csproj
```diff
@@ -1,9 +1,14 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>net7.0</TargetFramework>
+		<TargetFramework>net8.0</TargetFramework>
 		<IsPackable>false</IsPackable>
-		<LangVersion>11.0</LangVersion>
+		<LangVersion>12.0</LangVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <Nullable>true</Nullable>
```
timmytaters, line 8: I think my previous comment may be wrong, when i downloaded it the nullability was reverted

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,37 @@
...
+    {
+        LogHelper(logger, LogLevel.Information, message, messages);
+    }
+
+    public static void Debug(this BaseLogger? logger, string message, params string[] messages)
+    {
+        LogHelper(logger, LogLevel.Debug, message, messages);
+    }
+
+    private static void LogHelper(BaseLogger? logger, LogLevel level, string message, params string[] messages)
```
timmytaters, line 27: Rather than a helper method, I did string.Format with all of the inputs for the message, and then it combines the messages automatically

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
...
+                File.Create(value).Dispose();
+            }
+        }
+    }
+    private readonly string? _FilePath;
+    public override void Log(LogLevel logLevel, string message)
+    {
+        string output = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt", CultureInfo.CurrentCulture);
+        output += " " + nameof(FileLogger) + " " + logLevel + ": " + message;
+        File.AppendAllText(_FilePath!, Environment.NewLine + output);
```
timmytaters, line 26: I think instead of doing Environment.NewLine you can concatenate a "\n", but this may be better

timmytaters (commented):
Placeholder

Logger/FileLogger.cs
Joshua-Lester3: I think Environment.NewLine gives a newline character regardless of the system you're on so I think it's more full-proof

Logger/BaseLoggerMixins.cs
Joshua-Lester3: Thanks for the tip! I fixed it, but kept the helper method in to reduce repeated lines :)

Logger.Tests/Logger.Tests.csproj
Joshua-Lester3: I guess for some reason "true" doesn't work, but "enable" does. I found that issue too. I'm not sure how it made it through the build. I fixed it!

Logger.Tests/FileLoggerTests.cs
Joshua-Lester3: I think the nature of AppendAllText (used in my FileLogger class) is that it appends the string at the end of the file. Here's the docs: https://learn.microsoft.com/en-us/dotnet/api/system.io.file.appendalltext?view=net-8.0 
> Given a string and a file path, this method opens the specified file, appends the string to the end of the file, and then closes the file.

Logger.Tests/BaseLoggerMixinsTests.cs
Joshua-Lester3: Thanks for commenting on this. I didn't really understand having the parameter array - I thought it was an array of strings to be added to the message. I get it now. I fixed it!

StrangeRanger:
@timmytaters Good job and thank you for putting notes next to the requirements that were missed. For the next assignment, please make sure to provide a comment in the code itself, where the missing requirement should go.

StrangeRanger:
@timmytaters  Make sure that all requirements that have been marked with an '❌', have their appropriate code commented on.

Example:

> Use the nameof() operator when identifying the class name to the logger ❌

If you think that the above requirement was not complete, make sure to comment on the code where `nameof()` should have been used.

However, I do appreciate that you at least added a small comment to one of the requirements that Joshua missed. This way they better know about what requirement needs to be worked on.

> The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ❌ - ConfigureFileLogger is not used within CreateLogger

PR #56, opened 2024-01-23
Assignment2 - Trevor Stewart

Tvor369:


Logger/LogFactory.cs
```diff
@@ -1,10 +1,36 @@
...
+    {
+        if (newFilePath != null && newFilePath != "")
+        {
+            _filePath = newFilePath;
+            return _filePath;
+        }
+        else
+        {
+            _filePath = null;//throw an exception?
+            throw new ArgumentNullException(_filePath,"File Path not set.");
```
Tfelchlin13, line 33: This is great you are prepared for all the cases

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,64 @@
...
+            throw new ArgumentNullException(nameof(logger), "Logger cannot be null.");
+        }
+
+        else
+        {
+            string informationMessage = string.Format(null, message, args);
+            logger.Log(LogLevel.Information, informationMessage);
+        }
+    }
+
```
Tfelchlin13, line 49: looks great i added a helper method and kept calling it it make it simpler but this works great.

Logger/Logger.csproj
```diff
@@ -1,8 +1,13 @@
...
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>11.0</LangVersion>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>12.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNetAnalyzers>true</EnableNetAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
Tfelchlin13, line 10: You got them all perfect!

Tfelchlin13 (commented):

The current date/time ✔
The log level ✔
The message ✔

Error, ✔
Warning, ✔
Information, and ✔
Debug. ✔

 Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.
Extra Credit


This code looks really good!

Logger/BaseLoggerMixins.cs
Tvor369: Great idea! I think that was one of the suggestions visual studio gave me, I'm just lazy

PR #62, opened 2024-01-23
Assignment 2 - Paulson Ray Thompson

Purplepigs0129:
Here it is hopefully done

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -31,8 +31,7 @@ public class FileLoggerTests
         MyFile.Close();
         _fileLogger!.ClassName = "Test";
         _fileLogger!.Log(LogLevel.Error, message);
-        //Had to subtract 1 second this might changed depending on how fast your machine is
-        DateTime dateTime = DateTime.Now.AddSeconds(-1);
+        DateTime dateTime = DateTime.Now;
         StreamReader FileReader = new(_Path);
         string MessageInFile = FileReader.ReadLine() ?? string.Empty;
         FileReader.Close();
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System;
+using System.IO;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Logger.Tests;
 
@@ -6,4 +9,37 @@ namespace Logger.Tests;
 public class FileLoggerTests
 {
 
+    private FileLogger? _fileLogger;
+    private readonly string _Path = "Text.txt";
+    [TestInitialize]
+    public void Constructor()
+    {
+        _fileLogger = new(_Path);
+    }
+
+    [TestMethod]
+    public void Log_AppendsMessage_Successful()
+    {
+        string message = "Hello Everybody!";
+       
+        if (File.Exists(_Path))
+        {
+            File.Delete(_Path);
+        }
+
+        FileStream MyFile = File.Create(_Path);
+        MyFile.Close();
+        _fileLogger!.ClassName = "Test";
+        _fileLogger!.Log(LogLevel.Error, message);
+        //Had to subtract 1 second this might changed depending on how fast your machine is
+        DateTime dateTime = DateTime.Now.AddSeconds(-1);
+        StreamReader FileReader = new(_Path);
+        string MessageInFile = FileReader.ReadLine() ?? string.Empty;
+        FileReader.Close();
+        string MessageToAppend = $"{dateTime} Test Error : {message}";
+
+        Assert.AreEqual(MessageToAppend, MessageInFile);
+
+        File.Delete(_Path);
+    }
 }
--- a/Logger.Tests/LogFactoryTests.cs
+++ b/Logger.Tests/LogFactoryTests.cs
@@ -1,9 +1,27 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System;
+using System.IO;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Logger.Tests;
 
 [TestClass]
 public class LogFactoryTests
 {
+    private LogFactory? _logFactory;
+    private BaseLogger? _logger;
+    [TestInitialize]
+    public void Constructor()
+    {
+        _logFactory = new();
+        _logFactory.ConfigureFileLogger("Test.txt");
+        _logger = _logFactory.CreateLogger("Test");
+    }
+
+    [TestMethod]
+    public void ConfigureFileLogger_GoodFilePath_Successful()
+    {
+        Assert.AreEqual("Test.txt",_logFactory!.GetFileName());
+    }
 
 }
--- a/Logger.Tests/LogFactoryTests.cs
+++ b/Logger.Tests/LogFactoryTests.cs
@@ -14,14 +14,20 @@ public class LogFactoryTests
     public void Constructor()
     {
         _logFactory = new();
-        _logFactory.ConfigureFileLogger("Test.txt");
-        _logger = _logFactory.CreateLogger("Test");
     }
 
     [TestMethod]
     public void ConfigureFileLogger_GoodFilePath_Successful()
     {
+        _logFactory!.ConfigureFileLogger("Test.txt");
+        _logger = _logFactory.CreateLogger("Test");
         Assert.AreEqual("Test.txt",_logFactory!.GetFileName());
+    }
+    [TestMethod]
+    public void CreateLogger_NoPathConfigured_ReturnsNull()
+    {
+        _logger = _logFactory!.CreateLogger("Test");
+        Assert.IsNull(_logger);
     }
 
 }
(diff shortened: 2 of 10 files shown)


Purplepigs0129:
Originally had a remove second to work on my machine if the test fails on yours just add it in appropriately using .addTime on fileloggertests.cs

BenjaminMichaelis:
Ignore the one specific failure on https://github.com/IntelliTect-Samples/EWU-CSCD371-2024-Winter/actions/runs/7621660789/job/20758379640?pr=62, not sure why a second action triggered.

Tvor369:
reviewing

Logger.Tests/Logger.Tests.csproj
```diff
@@ -1,9 +1,14 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>net7.0</TargetFramework>
+		<TargetFramework>net8.0</TargetFramework>
 		<IsPackable>false</IsPackable>
-		<LangVersion>11.0</LangVersion>
+    <LangVersion>11.0</LangVersion>
```
Tvor369, line 6: Latest LangVersion of C# is 12.0 not 11.0

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,8 @@
 
 public abstract class BaseLogger
 {
-    public abstract void Log(LogLevel logLevel, string message);
+    public abstract void Log(LogLevel LogLevel, string Message);
```
Tvor369, line 5: Don't we use camelCase for variable names? Maybe I'm mistaken here.
```suggestion
    public abstract void Log(LogLevel logLevel, string message);
```

Logger/LogFactory.cs
```diff
@@ -2,9 +2,25 @@
...
+    public BaseLogger? CreateLogger (string className)
+    {
+        if (className == nameof(FileLogger))
+        {
+            FileLogger FileLogger = new(_loggerPath!) { ClassName = className };
+            return FileLogger;
+        } else
+        {
+            return null;
+        }
```
Tvor369, line 16: Definitely not wrong but you could consider removing the else and just have return null by itself,  if the if statement didn't trigger then we wouldn't get that earlier return FileLogger and would instead be executing whatever is after the if statement anyway.
```suggestion
        }
        
        return null;
        
```

Logger/Logger.csproj
```diff
@@ -1,8 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>11.0</LangVersion>
+    <TargetFramework>net7.0</TargetFramework>
```
Tvor369, line 4: Should be net8.0
```suggestion
    <TargetFramework>net8.0</TargetFramework>
```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,9 +1,27 @@
...
+        _logFactory = new();
+        _logFactory.ConfigureFileLogger("Test.txt");
+        _logger = _logFactory.CreateLogger("Test");
+    }
+
+    [TestMethod]
+    public void ConfigureFileLogger_GoodFilePath_Successful()
+    {
+        Assert.AreEqual("Test.txt",_logFactory!.GetFileName());
+    }
```
Tvor369, line 26: I think you should add a test for the case when CreateLogger returns null when the filePath isn't set with ConfigureFileLogger.

Tvor369:
The current date/time ✔
The log level ✔
The message ✔
If the file logger has not been configured in the LogFactory, its CreateLogger method should return null. ✔
Error, ✔
Warning, ✔
Information, and ✔
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌✔

Overall good job, I can tell you tried to set the LangVersion and TargetFramework correctly as the project file for your logger.tests had the right TargetFramwork but not LangVersion, and your logger project file had the opposite issue. The other thing is I think you need to add a test to LogFactory for returning null when no filePath is set.

Logger/Logger.csproj
Purplepigs0129: Yes it should originally had these set to the proper ones might have gotten mixed up in the process of trying to get the checks to pass.

Logger/LogFactory.cs
Purplepigs0129: Ahh thanks for the catch and explanation will fix

Logger/BaseLogger.cs
Purplepigs0129: Yes a little mistake I found to late on Monday to fix had it in my plans thanks for the catch

Logger.Tests/Logger.Tests.csproj
Purplepigs0129: correct got mixed up int eh checks earlier will be fixed

Logger.Tests/LogFactoryTests.cs
Purplepigs0129: Will do just ran out of time on monday

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,73 @@
-﻿namespace Logger;
+﻿using System;
+using System.Runtime.CompilerServices;
+
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
+    public static void Error(this BaseLogger? logger, string message, params object[] args)
```
BenjaminMichaelis, line 8: Why is logger nullable? I expect to call this message, a logger is needed.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,44 @@
...
+        string message = "Hello Everybody!";
+       
+        if (File.Exists(_Path))
+        {
+            File.Delete(_Path);
+        }
+
+        FileStream MyFile = File.Create(_Path);
+        MyFile.Close();
+        _fileLogger!.ClassName = "Test";
```
BenjaminMichaelis, line 32: Why are you using a `!` here instead of checking if the logger is in fact not null?

Logger.Tests/Logger.Tests.csproj
```diff
@@ -1,9 +1,14 @@
...
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <Nullable>enable</Nullable>
```
BenjaminMichaelis, line 11:
```suggestion
		<LangVersion>11.0</LangVersion>
		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
		<EnableNETAnalyzers>true</EnableNETAnalyzers>
		<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
		<Nullable>enable</Nullable>
```
nit: spacing

Logger.Tests/Logger.Tests.csproj
BenjaminMichaelis, line 6: Latest is 12.0

Logger/BaseLogger.cs
```diff
@@ -3,5 +3,7 @@
 public abstract class BaseLogger
 {
     public abstract void Log(LogLevel logLevel, string message);
+
+    public string? ClassName { get; set; }
```
BenjaminMichaelis, line 7: Why is this nullable?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text;
+
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+    private readonly string _Path;
+    public FileLogger(string filePath)
+    {
+        this._Path = filePath;
```
BenjaminMichaelis, line 13:
```suggestion
        _Path = filePath;
```
why is this used

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+    private readonly string _Path;
```
BenjaminMichaelis, line 10: This should be a property, not a field

Logger/LogFactory.cs
```diff
@@ -2,9 +2,25 @@
 
 public class LogFactory
 {
-    public BaseLogger CreateLogger(string className)
-    {
+    private string? _loggerPath;
 
+    public BaseLogger? CreateLogger (string className)
+    {
+        if (className == nameof(FileLogger))
```
BenjaminMichaelis, line 9: This is weird.... How is this supposed to take a class name with this check?

Logger/LogFactory.cs
```diff
@@ -2,9 +2,25 @@
...
+            FileLogger FileLogger = new(_loggerPath!) { ClassName = className };
+            return FileLogger;
+        }
+        
         return null;
+        
+    }
+    public void ConfigureFileLogger(string FilePath)
+    {
+        this._loggerPath = FilePath;
```
BenjaminMichaelis, line 20: `this` should not be needed here

PR #72, opened 2024-01-23
Assignment2 Andree Hadi

HadiM2:


HadiM2:
Why is build failing?

EthanAGi:
Reviewing

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.IO;
+
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+
+    string filePath;
```
EthanAGi, line 9: changing this to a private, readonly field will fix this error.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,52 @@
...
+
+    public static void Debug(this BaseLogger baseLogger, string message, params object[] arguments)
+    {
+        if (baseLogger == null)
+        {
+            throw new System.ArgumentNullException(nameof(baseLogger));
+        }
+        else
+        {
+            baseLogger.Log(LogLevel.Debug, string.Format(message, arguments));
```
EthanAGi, line 49: changing this to an IFormatProvider where indicated will elminate this error. What it is worried about is that this string needs to be formatted for different machines. A recommendation from me would be CultureInfo.InvariantCulture, as this fomrats strings to the users machine.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,32 @@
...
+    public BaseLogger? CreateLogger(string className)
+    {
+        ConfigureFileLogger(filePath);
+        if(filePath == null)
+        {
+            return null;
+        }
+        else 
+        {
+            FileLogger fileLogger = new FileLogger(filePath) { ClassName = className};
```
EthanAGi, line 16: Here, if we use "FileLogger filelogger = new(filePath)" here, the Error will be removed. This is because in C#, Not using the class type next to new is the standard.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+
+    string filePath;
+    public FileLogger(string filePath)
+    {
+        this.filePath = filePath;
+    }
+    public override void Log(LogLevel logLevel, string message)
+    {
+        string dateTime = System.DateTime.Now.ToString("M/d/yyyy hh:mm:ss tt");
```
EthanAGi, line 16: We can replace this entire String by placing "DateTime.Now" into our call for logEntry. 
string logEntry = $"{DateTime.Now} {nameof(this.ClassName)} {logLevel}: {message}";
That worked form me, but I'm not too sure if that will directly fix the error.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,52 @@
...
+
+    public static void Information(this BaseLogger baseLogger, string message, params object[] arguments)
+    {
+        if (baseLogger == null)
+        {
+            throw new System.ArgumentNullException(nameof(baseLogger));
+        }
+        else
+        {
+            baseLogger.Log(LogLevel.Information, string.Format(message, arguments));
```
EthanAGi, line 37: Placing the IFormatProvider into this Line will also fix it.

EthanAGi (changes requested):
- There is an existing `BaseLogger` class. It needs an **auto property** to hold a class name. This property should be set in the `LogFactory` using an **object initializer**. ✔
- Create a `FileLogger` that derives from `BaseLogger`. It should take in a path to a file to write the log message to. When its `Log` method is called, it should **append** messages on their own line in the file. The output should include all of the following:
  - The current date/time ❌ <- The fix for this is indicated in the comments!
  - The name of the class that created the logger ✔
  - The log level ✔
  - The message ✔
  - The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
- The `LogFactory` should be updated with a new method `ConfigureFileLogger`. This should take in a file path and store it in a **private member**. It should use this when instantiating a new `FileLogger` in its `CreateLogger` method. ✔
- If the file logger has not be configured in the `LogFactory`, its `CreateLogger` method should return `null`. ✔
- Inside of `BaseLoggerMixins` implement **extension methods** on `BaseLogger` for
  - `Error`, ✔
  - `Warning`, ✔
  - `Information`, and ✔
  - `Debug`. ✔
  Each of these methods should take in a `string` for the message, as well as a **parameter array** of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the `BaseLogger.Log` method, by automatically supplying the appropriate `LogLevel`. These methods should throw an exception if the `BaseLogger` parameter is null. There are a couple example unit tests to get you started.
- Use the nameof() operator when identifying the class name to the logger ✔
- Ensure you turn on Warnings as Errors (TreatWarningsAsErrors) ✔
- Ensure that you turn on code analysis (EnableNETAnalyzers) ✔
- Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
- Ensure that you turn on EnforceCodeStyleInBuild ✔
- Turn on Nullability (`Nullable`) ✔

Nicely done! I can tell you get what's going on here, and with just a few tweaks, I think you'll be there! Well done!

Logger/BaseLoggerMixins.cs
HadiM2: Thank you that worked!

Logger/LogFactory.cs
HadiM2: OHHH it was because of the treat warnings as error, good catch!

Logger/FileLogger.cs
Aramirez61: Thank you, that did it and fixed the error we were getting.

Logger/BaseLoggerMixins.cs
Aramirez61: Thank you, we used string.Format(CultureInfo.InvariantCulture, message, args) and that fixed it

.DS_Store
BenjaminMichaelis, line 1: I would disclude these files from your commit's by putting them in your .gitignore

Logger.Tests/Logger.Tests.csproj
```diff
@@ -1,20 +1,21 @@
...
-
-</Project>
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>net8.0</TargetFramework>
+		<IsPackable>false</IsPackable>
+		<LangVersion>12.0</LangVersion>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
```
BenjaminMichaelis, line 8: It looks like you are missing a few elements here. Ex: TreatWarningsAsErrors

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,7 @@
 
 public abstract class BaseLogger
 {
+    public string? ClassName {get; set;}
```
BenjaminMichaelis, line 5: Why is this nullable?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.IO;
+
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+
+    private readonly string filePath;
```
BenjaminMichaelis, line 9: This should be a property

Logger/FileLogger.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+
+    private readonly string filePath;
+    public FileLogger(string filePath)
+    {
+        this.filePath = filePath;
```
BenjaminMichaelis, line 12: If you have to use `this` it indicates you are missing a naming convention. The field you are referencing should have a leading `_`

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+
+public class LogFactory
+{
+    private string? filePath;
+
+    //public FileName { get; private set; };
+
+    public BaseLogger? CreateLogger(string className)
+    {
+        ConfigureFileLogger(filePath);
```
BenjaminMichaelis, line 11: The logger should either be configured or not at this point. I don't see the logic in calling this method

Logger/Logger.csproj
```diff
@@ -1,8 +1,13 @@
...
-  </PropertyGroup>
-
-</Project>
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <LangVersion>12.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <NetAnalyzers>true</NetAnalyzers>
```
BenjaminMichaelis, line 7: This should be `EnableNETAnalyzers`

PR #77, opened 2024-01-23
Assignment2 - Conner Hanson

connerhanson:
I think since I removed (rm -r) and re-cloned my repo early on, it shows as if I remade the whole file for each file. I tried to fix this but I'm not exactly sure how I would since I did this at the beginning of the assignment.  Let me know if you have any questions while reviewing. 

After asking for help I think I've mostly fixed the above stated problem.  If it isn't looking right to you, set the Diff View to "Unified" and select "Hide whitespace".

### DIFF
--- a/Logger.Tests/Logger.Tests.csproj
+++ b/Logger.Tests/Logger.Tests.csproj
@@ -1,9 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>net7.0</TargetFramework>
+		<TargetFramework>net8.0</TargetFramework>
 		<IsPackable>false</IsPackable>
-		<LangVersion>11.0</LangVersion>
+    <LangVersion>12.0</LangVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <Nullable>enable</Nullable>
 	</PropertyGroup>
 
 	<ItemGroup>
--- a/Logger.Tests/Logger.Tests.csproj
+++ b/Logger.Tests/Logger.Tests.csproj
@@ -3,13 +3,13 @@
 	<PropertyGroup>
 		<TargetFramework>net8.0</TargetFramework>
 		<IsPackable>false</IsPackable>
-    <LangVersion>12.0</LangVersion>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
-    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
-    <Nullable>enable</Nullable>
-	</PropertyGroup>
+        <LangVersion>12.0</LangVersion>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <EnableNETAnalyzers>true</EnableNETAnalyzers>
+        <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+        <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+        <Nullable>enable</Nullable>
+      </PropertyGroup>
 
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -7,7 +7,7 @@ namespace Logger;
 public abstract class BaseLogger
 {
     //Auto Property
-    public string? ClassName { get; set; }
+    public string ClassName { get; set; }
 
     public abstract void Log(LogLevel logLevel, string message);
 }
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -1,7 +1,41 @@
-﻿namespace Logger;
+﻿using System;
+using System.IO;
+
+namespace Logger;
+#nullable enable
 
 public abstract class BaseLogger
 {
+    //Auto Property
+    public string? ClassName { get; set; }
+
     public abstract void Log(LogLevel logLevel, string message);
 }
 
+
+//FileLogger || Implementation of BaseLogger
+public class FileLogger : BaseLogger
+{
+    // create initializer for FileLogger
+    public string _fileName;
+    public FileLogger(string fileName)
+    {
+        _fileName = fileName;
+    }
+
+    //current date/time 
+    string dateTime = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt");
+
+    //create variable for class name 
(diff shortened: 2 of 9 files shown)


Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,61 @@
...
+        {
+            throw new System.ArgumentNullException(nameof(logger));
+        }
+        
+        //format string to return, use string.format so the argument is placed into the {0} spot
+        string formattedString = string.Format(CultureInfo.CurrentCulture, message, arguments);
+        
+        //Call log with specified loglevel for method and string
+        logger.Log(LogLevel.Error, formattedString);
+    }
```
omar5o9, line 21:
```suggestion
        if(logger is null)
        {
            throw new System.ArgumentNullException(nameof(logger));
        }
        
        //format string to return, use string.format so the argument is placed into the {0} spot
        string formattedString = string.Format(CultureInfo.CurrentCulture, message, arguments);
        
        //Call log with specified loglevel for method and string
        logger.Log(LogLevel.Error, formattedString);
    }

```
For these methods, instead checking for null and formatting each one, why not try to create a helper method that does all of these things to save you some code by calling the helper in each one instead.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,35 @@
...
+﻿namespace Logger;
+#nullable enable
+
+public class LogFactory
+{
+    //create nullable path string
+    private string? path;   
+
+    //Configure given filePath as path
+    public void ConfigureFileLogger(string filePath)
```
omar5o9, line 10: This method should be used in the createLogger method
```suggestion
    public void ConfigureFileLogger(string filePath)

```

omar5o9 (commented):
The current date/time ✔
The log level ✔
The message ✔
The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ❌ (Didn't use use the configure in the CreateLogger method)
Error, ✔
Warning, ✔
Information, and ✔

Logger/BaseLoggerMixins.cs
connerhanson: I initially meant to do that, but it looks like I forgot to implement it thanks for letting me know!

Logger/LogFactory.cs
connerhanson: Thanks for pointing that out should be fixed now.

KairyuSen:
The current date/time ✔
The log level ✔
The message ✔
Error, ✔
Warning, ✔
Information, and ✔

Looking at your code it looks like you deleted what was initially there in the docs then retyped it, is this the case and if so why? Just curious to the thought process

connerhanson:
@KairyuSen 
As I said above I rm -r'd my initial cloned repo and recloned it to restart. I think the .git file didn't recognize that and shows it as all new code.

Logger.Tests/Logger.Tests.csproj
```diff
@@ -1,9 +1,14 @@
...
+		<TargetFramework>net8.0</TargetFramework>
 		<IsPackable>false</IsPackable>
-		<LangVersion>11.0</LangVersion>
+    <LangVersion>12.0</LangVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <Nullable>enable</Nullable>
 	</PropertyGroup>
```
BenjaminMichaelis, line 12:
```suggestion
        <LangVersion>12.0</LangVersion>
        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
        <EnableNETAnalyzers>true</EnableNETAnalyzers>
        <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
        <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
        <Nullable>enable</Nullable>
	</PropertyGroup>
```
spacing should be consistent

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,43 @@
...
-{
-    public abstract void Log(LogLevel logLevel, string message);
-}
-
+﻿using System;
+using System.IO;
+using System.Globalization;
+
+namespace Logger;
+#nullable enable
```
BenjaminMichaelis, line 6: Why is these needed?

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,43 @@
...
+
+public abstract class BaseLogger
+{
+    //Auto Property
+    public string? ClassName { get; set; }
```
BenjaminMichaelis, line 11: Why is this nullable?

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,43 @@
...
+    {
+        _fileName = fileName;
+    }
+
+
+    //current date/time 
+    private string dateTime = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt", CultureInfo.CurrentCulture);
+
+    //create variable for class name 
+    private string className = nameof(FileLogger);
```
BenjaminMichaelis, line 32: Why is this private? We should be using properties instead of fields. Also we should be able to set the class.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,45 @@
...
-
-}
+﻿namespace Logger;
+using System.Globalization;
+#nullable enable
+
+public static class BaseLoggerMixins
+{
+
+    public static void Error(this BaseLogger? logger, string message, params object[] arguments)
```
BenjaminMichaelis, line 8: Why is logger nullable? I expect to call this message, a logger is needed.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+        //If given null path return null
+        if(path == null)
+        {
+            return null;
+        }
+
+        //If its not null then create new FileLogger and set class name and return the logger
+        else
+        {
+            FileLogger fileLogger = new(ConfigureFileLogger(path));
```
BenjaminMichaelis, line 29: Not sure why you need to call ConfigureFileLogger here or what that will do

Logger/BaseLogger.cs
connerhanson: I had put these in before enabling nullable in the csproj files.  I'll make sure when refactoring on the next assignment, to take out redundant or unused code before the due date.

Logger/LogFactory.cs
connerhanson: I put this here to resolve: 

The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ❌ (Didn't use use the configure in the CreateLogger method)

I realize that was not the correct way to implement this step, and that what I had before worked just as well.

Logger/BaseLogger.cs
connerhanson: For some reason when removing the nullability the build fails on the pull request I believe this is why I added it. Will investigate more tommorrow.

Logger/BaseLoggerMixins.cs
connerhanson: I did this to resolve: 

"These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started."

I think this is correct since it says if the baseLogger is null throw an exception and you would need to input null for this to happen, though I maybe miss understanding.  I will ask in class tommorrow.

PR #84, opened 2024-01-25
Assignment2 - Ethan Florko

EthanFlorko:


Aramirez61:
I'll review this.

Logger.Tests/Logger.Tests.csproj
```diff
@@ -4,6 +4,17 @@
 		<TargetFramework>net7.0</TargetFramework>
```
Aramirez61, line 4: May want to change target framework to net 8.0 and lang version to 12.0.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,38 @@
...
-public class FileLoggerTests
+namespace Logger.Tests
 {
 
-}
+    [TestClass]
+    public class FileLoggerTests
+    {
+        [TestMethod]
+        public void Log_ValidInput_LogsMessageToFile()
```
Aramirez61, line 13: Good use of following the unit test naming conventions.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -5,5 +5,32 @@ namespace Logger.Tests;
...
+        //Act
+        var logger = logFactory.CreateLogger("ExampleClass");
+        
+        //Assert
+        Assert.IsNotNull(logger);
+        Assert.IsInstanceOfType(logger, typeof(FileLogger));
+    }
+    
+    [TestMethod]
+    public void CreateLogger_WithoutConfiguredFileLogger_ReturnsNull()
```
Aramirez61, line 25: Good testing of covering cases when ConfigureFileLogger is not used.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,49 @@
...
-public static class BaseLoggerMixins
+namespace Logger
 {
 
-}
+    public static class BaseLoggerMixins
+    {
+        public static void Error(this BaseLogger logger, string message, params object[] args)
+        {
+            LogWithLevel(logger, LogLevel.Error, message, args);
```
Aramirez61, line 13: I think all of these LogWithLevel calls should be replaced with baseLogger.Log() method calls. I am not able to see where the LogWithLevel method fits into BaseLoggerMixins class.

Aramirez61 (commented):

The current date/time ✔
The log level ✔
The message ✔
Error, ❌
Warning, ❌
Information, and ❌
Debug. ❌ Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.
-
(NOTE: extension methods are implemented here but need to be using the BaseLogger.log, which is not currently being used in this BaseLoggerMixins class).

Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌
Extra Credit

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class FileLogger : BaseLogger
+    {
+        //Private mem to store file path
+        private readonly string filePath;
+
+        //Constructor
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
```
BenjaminMichaelis, line 14: `this` is not needed if the field is named properly with a leading underscore

PR #85, opened 2024-01-26
Assignment3 Matt Adams and Rachael DeBoer

rkdeboer:
@MattA41 and @rkdeboer


Joshua-Lester3:
started review

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+
+namespace CanHazFunny; 
+    public interface IOutPutJoke
+    {
+        public void Print(String Joke);
+    }
+    public class Jester(IJokeServiceInterface jokeService, IOutPutJoke output)
+{
+    private readonly IJokeServiceInterface jokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
```
Joshua-Lester3, line 15: private fields need to have an underscore at the start of their name. One should also use a property here.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly IOutPutJoke output = output ?? throw new ArgumentNullException(nameof(output));
```
Joshua-Lester3, line 16: see comment above (add underscore and implement as property)

CanHazFunny/CanHazFunny/OutputService.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+    public class OutputService : IOutPutJoke
+    {
+        public void Print(string Joke)
```
Joshua-Lester3, line 11: Would we want to throw an exception if null here? (I'm not actually sure.)

Joshua-Lester3 (commented):

Fundamentals
Be sure you enable:
For this assignment, always use Assert.Equal<T>() (the generic version) ❌ Did not use generics throughout pr (first instance I found was in JesterTests.cs on line 25)
All of the above should be unit tested ❌ I may be a little picky here, but I feel like the jokeservice could be unit-tested to assert it's not returning a null or empty string.

Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ❌ no unit test for writing joke to screen
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. ❌ doesn't retrieve jokes using JSON

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CanHazFunny; 
+    public interface IOutPutJoke
+    {
+        public void Print(String Joke);
+    }
```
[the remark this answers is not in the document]
rkdeboer: I was confused on how the interfaces were supposed to be laid out when I started the project. I updated this as it makes more sense with current implementation thanks!

CanHazFunny/CanHazFunny.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
```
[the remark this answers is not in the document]
rkdeboer: I will say I was a bit confused about this since I was under the impression earlier that we would need 100% code coverage but found out later it was just for jester. I removed the class

CanHazFunny/CanHazFunny.Tests/OutputServiceTests.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace CanHazFunny.Tests;
+    public class OutputServiceTests
```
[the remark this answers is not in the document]
rkdeboer: This comment was a lifesaver thank you so much!!

CanHazFunny/CanHazFunny/Jester.cs
rkdeboer: I am a little confused what you mean here. Could you elaborate?

CanHazFunny/CanHazFunny/OutputService.cs
rkdeboer: Its not a bad idea. I think it would depend on how its used but I added it anyway. Its good to check for those things!

CanHazFunny/CanHazFunny/IOutPutJoke.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IOutPutJoke
+{
+    public void Print(String Joke);
```
BenjaminMichaelis, line 11:
```suggestion
    public void Print(string Joke);
```
`string` preferred over `String`

CanHazFunny/CanHazFunny/Jester.cs
BenjaminMichaelis: Good suggestion

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,13 +2,14 @@
 
 namespace CanHazFunny;
 
-public class JokeService
+public class JokeService : IJokeServiceInterface
```
BenjaminMichaelis, line 5:
```suggestion
public class JokeService : IJokeService
```
The `I` prefix indicates that it's an interface. The word is unnecessary/redundant

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,13 +2,14 @@
...
 {
     private HttpClient HttpClient { get; } = new();
 
     public string GetJoke()
     {
-        string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+        HttpResponseMessage json = HttpClient.GetAsync("https://geek-jokes.sameerkumar.website/api").Result;
+        string joke = json.Content.ReadAsStringAsync().Result;
```
BenjaminMichaelis, line 12:
```suggestion
        string json = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
        var joke = JsonSerializer.Deserialize<JokeInformation>(json);
```
I think I would opt to use a JsonSerializer here

PR #86, opened 2024-01-27
Assignment3 - Cole Stewart

colestewart12:


Purplepigs0129:
Reviewing

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,8 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
-
+    <TargetFramework>net8.0</TargetFramework>
+	<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+	<EnableNETAnalyzers>True</EnableNETAnalyzers>
+	<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
```
Purplepigs0129, line 7:
```
```
Missing Language Version

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,20 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
+	<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+	<EnableNETAnalyzers>True</EnableNETAnalyzers>
+	<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
```
Purplepigs0129, line 8:
```
```
Missing Language Version

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,22 @@
+﻿
+
+using System;
+
+namespace CanHazFunny;
+
+public class Jester(IJokeOutput jokeOutput, IJokeService jokeService)
+{
+    private readonly IJokeService _jokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
```
Purplepigs0129, line 9: I believe you don't need the readonly on either the jokeservice or the jokeoutput believe you only need the private unless you have a specific reason for them

CanHazFunny/CanHazFunny/Program.cs
```diff
@@ -1,10 +1,13 @@
...
 {
-    static void Main(string[] args)
+    public static void Main(string[] args)
     {
-        //Feel free to use your own setup here - this is just provided as an example
-        //new Jester(new SomeReallyCoolOutputClass(), new SomeJokeServiceClass()).TellJoke();
+
+        var OutputInstance = new JokeOutput();
+        var ServiceInstance = new JokeService();
+
```
Purplepigs0129, line 10: I don't see the reason for needing to initialize the output and the service when you could do it in the new Jester.
```suggestion
new Jester(new JokeOutput(), new JokeService()).TellJoke();
```

CanHazFunny/CanHazFunny/Program.cs
```diff
@@ -1,10 +1,13 @@
 ﻿namespace CanHazFunny;
 
-class Program
+public  sealed class Program
```
Purplepigs0129, line 3: I also had the suggestion of having this sealed. I don't know if it should, but the public is there which I think it does need. Does the sealed server a purpose here or just extra?

CanHazFunny/CanHazFunny.Tests/ProgramTests.cs
Purplepigs0129, line 1: Good job going for the extra credit and it looks corrrect

Purplepigs0129 (commented):
Fundamentals
Be sure you enable:
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌✔
Choose simplicity over complexity ❌✔
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.

Both your csproj files was missing language version but the other ones were still there. I think your test for program is the extra credit it doesn't say implicitly whether its that file or the Output file but from what your testing it looks correctly attempted. It seems you did try to keep your code simple however there were a couple of lines I saw that could be updated to be more simple potentially unless its required for your tests that is why I left it with the X and check mark. The same thing goes for LangVerson and TargetFramework because you had the TargetFramework but not the LangVersion. Good Work!

CanHazFunny/CanHazFunny/Program.cs
```diff
@@ -1,10 +1,13 @@
...
-    static void Main(string[] args)
+    public static void Main(string[] args)
     {
-        //Feel free to use your own setup here - this is just provided as an example
-        //new Jester(new SomeReallyCoolOutputClass(), new SomeJokeServiceClass()).TellJoke();
+
+        var OutputInstance = new JokeOutput();
+        var ServiceInstance = new JokeService();
+new Jester(new JokeOutput(), new JokeService()).TellJoke();
+        new Jester(OutputInstance, ServiceInstance).TellJoke();
```
colestewart12, line 11:
```suggestion
        new Jester(new JokeOutput(), new JokeService()).TellJoke();
```

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,8 +1,12 @@
...
+	<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<Nullable>enable</Nullable>
```
colestewart12, line 9:
```suggestion
    <TargetFramework>net8.0</TargetFramework>
    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
    <EnableNETAnalyzers>True</EnableNETAnalyzers>
    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
    <Nullable>enable</Nullable>
    <LangVersion>12.0</LangVersion>
```

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,20 @@
...
+	<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<Nullable>enable</Nullable>
```
colestewart12, line 10:
```suggestion
    <TargetFramework>net8.0</TargetFramework>
    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
    <EnableNETAnalyzers>True</EnableNETAnalyzers>
    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
    <Nullable>enable</Nullable>
    <LangVersion>12.0</LangVersion>
```

CanHazFunny/CanHazFunny/Jester.cs
colestewart12: I was thinking that by making them readonly they would become immutable and since we aren't chnaging anything about them it was better for security reasons.

CanHazFunny/CanHazFunny/Program.cs
colestewart12: The `sealed` makes the class unable to be extended or used for inheritance. I received a warning in VS and decided to add it since I felt there was never an instance where I would want to extend or modify the Program class

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,8 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
-
+    <TargetFramework>net8.0</TargetFramework>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>True</EnableNETAnalyzers>
```
BenjaminMichaelis, line 6:
```suggestion
    <EnableNETAnalyzers>true</EnableNETAnalyzers>
```
nit: casing

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,57 @@
...
 public class JesterTests
 {
+        [Fact]
+        public void TellJoke_NormalData_SuccessfulReturn()
+        {
+            //Arrange
+            var mockJokeOutput = new Mock<IJokeOutput>();
+            var mockJokeService = new Mock<IJokeService>();
+
+            mockJokeService.SetupSequence(x => x.GetJoke()).Returns("Normal joke");
```
BenjaminMichaelis, line 17:
```suggestion
            mockJokeService.Setup(x => x.GetJoke()).Returns("Normal joke");
```
`SetupSequence` is not needed here.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,57 @@
...
+
+            var jester = new Jester(mockJokeOutput.Object, mockJokeService.Object);
+
+            //Act
+            jester.TellJoke();
+
+            //Assert
+            Assert.Equal("Normal joke", mockJokeOutput.Invocations[0].Arguments[0] as string);
```
BenjaminMichaelis, line 25: Not sure how I feel about this... This "works" but is way more difficult than it needs to be.

1. A possible better way: `mockJokeOutput.Verify(m => m.WriteOutput("Normal joke"), Times.Once(), "The output was not written correctly");`
2. marking the Setup and Verifable is another option. This can be done by `Setup` instead of `SetupSequence`. I assume `SetupSequence` is a result of copy and pasting from other places

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,57 @@
...
+            mockJokeService.SetupSequence(x => x.GetJoke())
+                .Returns("Chuck Norris joke")
+                .Returns("Normal joke");
+            var jester = new Jester(mockJokeOutput.Object, mockJokeService.Object);
+
+            //Act
+            jester.TellJoke();
+
+            //Assert
+            Assert.Equal("Normal joke", mockJokeOutput.Invocations[0].Arguments[0] as string);
```
BenjaminMichaelis, line 45: Same comment as before, you are strong-arming Moq instead of letting it do the hard work for you.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,57 @@
...
+        }
+
+        [Fact]
+        public void TellJoke_NullJokeOutput_ThrowsException()
```
BenjaminMichaelis, line 49: What if mockJokeService is null?

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,21 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>True</EnableNETAnalyzers>
```
BenjaminMichaelis, line 7:
```suggestion
    <EnableNETAnalyzers>true</EnableNETAnalyzers>
```
nit: Casing

CanHazFunny/CanHazFunny/Jester.cs
BenjaminMichaelis: `readonly` doesn't hurt in this case

CanHazFunny/CanHazFunny/Program.cs
BenjaminMichaelis: I would generally prefer `static` over sealed I think

PR #90, opened 2024-01-29
Assignment3 - Alexa Darrington and Johanne McClenahan

JohanneJayde:
Alexa's Github - @alyxadarr
Johanne's Github - @JohanneJayde


rkdeboer:
Starting review

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,14 +1,20 @@
...
-
-    public string GetJoke()
+    public class JokeService : IJokeService
     {
-        string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-        return joke;
+        private HttpClient HttpClient { get; } = new();
+        public string GetJoke()
+        {
+            HttpResponseMessage json = HttpClient.GetAsync("https://geek-jokes.sameerkumar.website/api").Result;
```
rkdeboer, line 14: Nice work making it to JSON!

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        }
+        public void TellJoke()
+        {
+            string joke;
+
+            do
+            {
+                joke = JokeService.GetJoke();
+            } while (joke.Contains("chuck norris", StringComparison.CurrentCultureIgnoreCase));
```
rkdeboer, line 25: Great use of the do-while loop

rkdeboer (commented):
Be sure you enable:

CanHazFunny/CanHazFunny/Jester.cs
alyxadarr: Thank you!

CanHazFunny/CanHazFunny/JokeService.cs
JohanneJayde: Thanks! I think I may have messed up on it so I'm going to refactor it more

JohanneJayde: I most definitely didn't do this correctly. I have now completed this feature. However, it's a rudimentary way of doing it. I feel like using Newtonsoft's JSON library would've been better, but I didn't know if we were allowed to use external libraries.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,14 +1,26 @@
...
-        return joke;
-    }
+        private HttpClient HttpClient { get; } = new();
+        public string GetJoke()
+        {
+            HttpResponseMessage json = HttpClient.GetAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
+
+            string joke = ParseJokeJSON(json.Content.ReadAsStringAsync().Result);
+
+            return joke;
```
BenjaminMichaelis, line 13: Not the worst way to parse json, but probably harder than it needs to be. I think something like this would be simpler
```suggestion
         string json = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
        var joke = JsonSerializer.Deserialize<JokeInformation>(json);
        if (joke != null && joke.Joke != null) return joke.Joke;
        throw new System.Exception("No joke received from API or joke was null");
```

PR #93, opened 2024-01-29
Assignment3 - Paulson Ray Thompson

Purplepigs0129:
First Attempt

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,7 +1,51 @@
 using Xunit;
+using Moq;
+using System;
 
 namespace CanHazFunny.Tests;
 
 public class JesterTests
 {
+    [Fact]
+    public void TellJoke_NormalTest_Successful()
+    {
+        var mockJokeToScreen = new Mock<IJokeToScreen>();
+        var mockJokeService = new Mock<IJokeService>();
+        mockJokeService.SetupSequence(x => x.GetJoke()).Returns("Test Joke");
+
+        var testJester = new Jester(mockJokeService.Object, mockJokeToScreen.Object);
+        testJester.TellJoke();
+
+        Assert.Equal("Test Joke", mockJokeToScreen.Invocations[0].Arguments[0]);
+    }
+
+    [Fact]
+    public void TellJoke_ChuckNorrisThenNormalTest_ReturnsNormalJoke()
+    {
+        var mockJokeToScreen = new Mock<IJokeToScreen>();
+        var mockJokeService = new Mock<IJokeService>();
+        mockJokeService.SetupSequence(x => x.GetJoke()).Returns("Chuck Norris").Returns("Test Joke");
+
+        var testJester = new Jester(mockJokeService.Object, mockJokeToScreen.Object);
+        testJester.TellJoke();
+
+        Assert.Equal("Test Joke", mockJokeToScreen.Invocations[0].Arguments[0]);
+    }
+
+    [Fact]
+    public void TellJoke_NullJokeToScreen_ThrowsException()
+    {
+        IJokeToScreen? nullJokeToScreen = null;
+        var mockJokeService = new Mock<IJokeService>();
+        Assert.Throws<ArgumentNullException>(() => new Jester(mockJokeService.Object, nullJokeToScreen!));
+    }
+
+    [Fact]
+    public void TellJoke_NullJokeService_ThrowsException()
+    {
+        IJokeService? nullJokeService = null;
+        var mockJokeToScreen = new Mock<IJokeToScreen>();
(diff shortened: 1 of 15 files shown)


Purplepigs0129:
If run locally might have popups when running tests should pass the tests still

montalcy:
Ill be reviewing this one!

CanHazFunny/CanHazFunny/JokeToScreen.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public class JokeToScreen : IJokeToScreen
+{
+    public void PrintJoke(string joke)
+    {
+        ArgumentNullException.ThrowIfNull(joke);
```
montalcy, line 13: I didnt think to check the joke for null

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public class Jester(IJokeService jokeService, IJokeToScreen jokeToScreen)
+{
+    private IJokeService _jokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
```
montalcy, line 11: This is a really efficient way to throw the exceptions, I did more than I could have done. Good Job!

CanHazFunny/CanHazFunny.Tests/JokeToScreenTests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace CanHazFunny.Tests;
+
+public class JokeToScreenTests
+{
+    [Fact]
+    public void PrintJoke_OutputsJoke_Successful()
+    {
+        JokeToScreen jokeToScreen = new();
+        StringWriter stringWriter = new(CultureInfo.InvariantCulture);
```
montalcy, line 19: For this test, are you creating your own console output to test that the consoleOutput throws out the same message? I couldnt wrap my head around how to test this.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,51 @@
 using Xunit;
+using Moq;
+using System;
 
 namespace CanHazFunny.Tests;
 
 public class JesterTests
```
montalcy, line 7: From the time I'm reviewing your code, It looks like your coverage is almost there!

montalcy (commented):
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ❌ ---Try and increasing your code coverage :)
Fundamentals
Be sure you enable:

Really good job, looks like you could increase your test coverage. Have you tried playing around with moq to test?

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
Purplepigs0129: Actually, it is, just the jester class needs to be at 90% I believe is what it says according to the Assignment. I don't believe the whole project needs to be at 90% just the jester but I added in the other tests since it just made sense to in my brain

CanHazFunny/CanHazFunny.Tests/JokeToScreenTests.cs
Purplepigs0129: Pretty much ya just create a sample output and then compare whether or not that matches what was written out to the console

CanHazFunny/CanHazFunny/Jester.cs
Purplepigs0129: I think you could do it with an ArgumentNullExcpetion.ThrowifNull(nameof(joke...)) but because I have them set up as variables in the class this just made more since that way it gets set properly and then if it happens to be trying to set to null then it throws the exception.

CanHazFunny/CanHazFunny/JokeToScreen.cs
Purplepigs0129: Thanks just made sense in my head to test it.

CanHazFunny/CanHazFunny.Tests/JokeToScreenTests.cs
```diff
@@ -0,0 +1,33 @@
...
+        Console.SetOut(stringWriter);
+        jokeToScreen.PrintJoke("Test Joke");
+        string consoleOutput = stringWriter!.ToString();
+
+        Assert.Equal("Test Joke" + Environment.NewLine, consoleOutput);
```
BenjaminMichaelis, line 24:
```suggestion
        Assert.NotNull(stringWriter);
        Assert.Equal("Test Joke" + Environment.NewLine, stringWriter.ToString());
```
This is simpler and better checks for `NotNull`

PR #99, opened 2024-01-30
Assignment 3 - Spencer West - Matthew Powers

Matthew-Powers:
@SaaWest @Matthew-Powers

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
+++ b/CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
@@ -21,6 +21,9 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="NUnit" Version="4.0.1" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+    <PackageReference Include="xunit.extensibility.core" Version="2.6.6" />
   </ItemGroup>
 
   <ItemGroup>
--- a/CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
+++ b/CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
@@ -1,7 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+	  <TargetFramework>net8.0</TargetFramework>
+	  <LanngVersion>12.0</LanngVersion>
+	  <Nullable>enable</Nullable>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	  <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
 
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -49,15 +49,13 @@ namespace CanHazFunny.Tests;
 
             var jester = new Jester(jokeService.Object, outPut.Object);
 
-            using (StringWriter stringWriter = new StringWriter())
-            {
-                Console.SetOut(stringWriter);
-                for(int i = 0; i < 10; i++)
-                {
-                    jester.TellJoke();
-                    string consoleOutput = stringWriter.ToString();
-                    Assert.That(!consoleOutput.Contains("Chuck Norris"));
-                }
-            }
+        using StringWriter stringWriter = new();
+        Console.SetOut(stringWriter);
+        for (int i = 0; i < 10; i++)
+        {
+            jester.TellJoke();
+            string consoleOutput = stringWriter.ToString();
+            Assert.That(!consoleOutput.Contains("Chuck Norris"));
         }
     }
+}
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,19 +1,15 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
-using Xunit;
 using NUnit.Framework;
 using Moq;
-using System.Collections.Generic;
 using Assert = NUnit.Framework.Assert;
 using System.IO;
-
 
 namespace CanHazFunny.Tests;
 [TestClass]
 public class JesterTests
 {
     [Test]
-    //[DataRow(new JokeService(), new TellJoke())]
     public void Constructor_JokerService_Null()
     {
         Assert.Throws<ArgumentNullException>(() => new Jester(null!, new OutputJoke()));
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -5,6 +5,7 @@ using NUnit.Framework;
 using Moq;
 using System.Collections.Generic;
 using Assert = NUnit.Framework.Assert;
+using System.IO;
 
 
 namespace CanHazFunny.Tests
@@ -18,12 +19,14 @@ namespace CanHazFunny.Tests
             Assert.Throws<ArgumentNullException>(() => new Jester(null!, new OutputJoke()));
 
         }
+
         [Test]
         public void Constructor_TellJoke_Null()
         {
             Assert.Throws<ArgumentNullException>(() => new Jester(new JokeService(), null!));
         }
-        [Fact]
+
+        [Test]
         public void TellJoke_JokeOutput()
         {
             var jokeService = new Mock<IJokerJokes>();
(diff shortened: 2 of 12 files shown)


Tvor369:
reviewing

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
```diff
@@ -0,0 +1,15 @@
+﻿using NUnit.Framework;
+using Assert = Microsoft.VisualStudio.TestTools.UnitTesting.Assert;
+
+namespace CanHazFunny.Tests;
+public class JokeServiceTests
+{
+    [Test]
+    public void GtJoke_ReturnJoke()
```
Tvor369, line 8:
```suggestion
    public void GetJoke_ReturnJoke_NotNull()
```
Just a small typo with GetJoke, I would also add what outcome you are wanting to the end of the name of the test method

CanHazFunny/CanHazFunny.Tests/OutputJokeTests.cs
```diff
@@ -0,0 +1,20 @@
...
+using NUnit.Framework;
+
+namespace CanHazFunny.Tests;
+public class OutputJokeTests
+{
+    [Test]
+    public void OutputJoke_toConsole()
+    {
+        var outputJoke = new OutputJoke();
+        var aJoke = "This is a joke";
```
Tvor369, line 13:
```suggestion
        OutputJoke outputJoke = new();
        string aJoke = "This is a joke";
```
Might be wrong about this but I think they wanted us to avoid using var when initializing variables and such, what you guys had originally still works though.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace CanHazFunny;
+public class Jester(IJokerJokes jokerJokes, IJokeOutput jokeOutput)
+{
+    public IJokeOutput IJokeOutput { get; set; } = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
```
Tvor369, line 6: good job using auto properties, and including the nameof() in the ArgumentNullException.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,17 @@
...
+    public IJokerJokes IJoker { get; set; } = jokerJokes ?? throw new ArgumentNullException(nameof(jokerJokes));
+
+    public void TellJoke()
+    {
+        string joke = IJoker.GetJoke();
+
+        while (joke.Contains("Chuck Norris"))
```
Tvor369, line 13: I used a do while loop here but this works as well.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,58 @@
...
+        var jester = new Jester(jokeService.Object, outPut.Object);
+
+        using StringWriter stringWriter = new();
+        Console.SetOut(stringWriter);
+        for (int i = 0; i < 10; i++)
+        {
+            jester.TellJoke();
+            string consoleOutput = stringWriter.ToString();
+            Assert.That(!consoleOutput.Contains("Chuck Norris"));
+        }
```
Tvor369, line 56:
```suggestion
    public void TellTenJokes_NoChuckNorris_Success()
    {
        var jokeService = new Mock<IJokerJokes>();
        jokeService.SetupSequence(s => s.GetJoke())
                .Returns("Chuck Norris")
                .Returns("Still Chuck Norris")
                .Returns("This is a joke");
        
        var outPut = new Mock<IJokeOutput>();

        var jester = new Jester(jokeService.Object, outPut.Object);

        using StringWriter stringWriter = new();
        Console.SetOut(stringWriter);

        jester.TellJoke();

//First joke TellJoke() will see is "Chuck Norris", so your TellJoke() method will call GetJoke() again
//The next joke GetJoke() Returns still contains Chuck so that while loop in TellJoke() will call GetJoke() again
//The last joke GetJoke() will return is good, so TellJoke() will exit that while loop

        string consoleOutput = stringWriter.ToString();
        Assert.That(!consoleOutput.Contains("Chuck Norris"));        

    }
```

Tvor369: The problem with the for loop is that it wont guarantee that your TellJoke method will ever receive a joke that contains Chuck Norris. In fact the way before was only running TellJoke with it recieving "This is a joke" ten times.
So instead we can either use a queue and have it dequeue in that .Returns(), or we can use SetupSequence with multiple Returns calls like this.

Tvor369:
### Assignment
For this assignment do the following:
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ❌
### Fundamentals
Be sure you enable:
### Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?✔?
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.

CanHazFunny/CanHazFunny/Jester.cs
SaaWest: thank you

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
SaaWest: Will fix

CanHazFunny/CanHazFunny.Tests/OutputJokeTests.cs
SaaWest: I'll change it I honestly don't recall

StrangeRanger: `var` should only be used when the type is obvious, such as the aJoke. Even then, it's usually recommended to specify the type over the use of `var`.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
StrangeRanger: Another option would be to do the following:

```csharp
do
{
    joke = JokeService.GetJoke();
} while (joke.Contains("Chuck Norris"));
```

This ensures that a new joke is retrieved, until the string "Chuck Norris" is no longer present in the joke.

Matthew-Powers: Implemented @Tvor369 's solution

CanHazFunny/CanHazFunny/Jester.cs
Matthew-Powers: Changed to a do while loop to get code coverage

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
Matthew-Powers: Fixed

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,7 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+	  <TargetFramework>net8.0</TargetFramework>
+	  <LanngVersion>12.0</LanngVersion>
```
BenjaminMichaelis, line 5:
```suggestion
	  <LangVersion>12.0</LangVersion>
```

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -15,6 +21,9 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="NUnit" Version="4.0.1" />
```
BenjaminMichaelis, line 24: Oh NUnit, out of curiosity, why do you prefer it?

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,89 @@
...
+    {
+        // Arrange
+        Jester jester = new(new JokeService(), new OutputJoke());
+        Mock<IJokerJokes> jokerJokesMock = new();
+
+        // Act
+        jester.IJoker = jokerJokesMock.Object;
+
+        // Assert
+        Assert.That(jokerJokesMock.Object == jester.IJoker);
```
BenjaminMichaelis, line 37: You are testing Moq more than anything valuable here

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,89 @@
...
+        Mock<IJokeOutput> outPut = new();
+        Jester jester = new(jokeService.Object, outPut.Object);
+        using StringWriter stringWriter = new();
+        Console.SetOut(stringWriter);
+        jester.TellJoke();
+        //First joke TellJoke() will see is "Chuck Norris", so your TellJoke() method will call GetJoke() again
+        //The next joke GetJoke() Returns still contains Chuck so that while loop in TellJoke() will call GetJoke() again
+        //The last joke GetJoke() will return is good, so TellJoke() will exit that while loop
+        string consoleOutput = stringWriter.ToString();
+        Assert.That(!consoleOutput.Contains("Chuck Norris"));
```
BenjaminMichaelis, line 87: I would do an invariant casing check generally

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace CanHazFunny;
+
+public interface IJokeOutput
+{
+    public void Output(string joke)
+    {
+
```
BenjaminMichaelis, line 7: You don't need curly braces, just a semi-colan like in IJokerJokes

SaaWest: sorry will remember that for next time

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
SaaWest: Reading on it now. So implementing something like this? 
`
string val = "Chuck Norris
string.Equals(val, "chuck norris", StringComparison.OrdinalIgnoreCase)
`

PR #103, opened 2024-01-30
Assignment3 - Conner Hanson

connerhanson:


alyxadarr:
reviewing :)

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    private IJokeService jokeService = jokeServiceInterface ?? throw new ArgumentNullException(nameof(jokeServiceInterface));
+
+    private IJokeOutput jokeOutput = jokeOutputInterface ?? throw new ArgumentNullException(nameof(jokeOutputInterface));
+
+    public void TellJoke()
+    {
+        string joke = jokeService.GetJoke();
+
+        while(joke.Contains("Chuck Norris"))
```
alyxadarr, line 16: Nice job using a while loop! I just wanted to share what me and my partner did. We used a do while loop and did 
do { 
joke=jokeService.GetJoke();
}while(joke.Contains("chuck norris");
Not a hard suggestion as your code is a good implementation as well.

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [Fact]
+    public void GetJoke_CreatesJoke_ReturnsNonNullJoke()
+    {
+        // Arrange
+        JokeService jokeService = new();
+        
+        // Act
+        string getJoke = jokeService.GetJoke();
+
```
alyxadarr, line 16: I like that you asserted that the joke wasn't null. You could also test for a Successful Return of a joke as well and create an expected joke. We used Mock for this and created an expectedJoke and then setup the sequence to return the expected joke and we used Assert.Equal on the actual and expected joke.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,67 @@
...
+        // Act
+        jesterTest.TellJoke();
+
+        // Assert with moq
+        moqJokeOutput.Verify(x => x.OutputJoke(testJoke));
+    }
+
+
+    [Fact]
+    public void OuputJoke_WithChuckNorrisThenWithout_Successful()
```
alyxadarr, line 49: I would change this test header to ``` OutputJoke_ChuckNorrisJoke_SuccessfulSkip()
```
just to make it clearer and shorter!

File: CanHazFunny/CanHazFunny/Jester.cs [hunk 4]
```diff
@@ -0,0 +1,26 @@
+using System;
+
+namespace CanHazFunny;
+
+
+public class Jester(IJokeService jokeServiceInterface, IJokeOutput jokeOutputInterface)
```
alyxadarr, line 6: Great job using the primary constructor for this!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,67 @@
...
+        // Arrange
+        var moqJokeService = new Mock<IJokeService>();
+        var moqJokeOutput = new Mock<IJokeOutput>();
+        string testJoke = "test joke";
+        moqJokeService.SetupSequence(x => x.GetJoke()).Returns(testJoke);
+        Jester jesterTest = new(moqJokeService.Object, moqJokeOutput.Object);
+        
+        // Act
+        jesterTest.TellJoke();
+
```
alyxadarr, line 42: Nice job using Mock. You could also test for how many times the GetJoke method skips the joke if it contains Chuck Norris. We setup the sequence to return more than one joke like this 
chuckJoke = "When God said,  Let there be light! Chuck Norris said, Say Please.";
noChuckJoke = "Why did the programmer break up with his girlfriend? She just didn't meet his conditional statements.";
 jokeMock.SetupSequence(jokeService => jokeService.GetJoke())
     .Returns(chuckJoke)
     .Returns(noChuckJoke);
and fed it chuck norris and non-chuck norris jokes and used Times.Exactly(1) in the Verify method to check. We also used Times.Once() which could be used to test the OutputToScreen method. Not a hard suggestion as you have the branch and line coverage but this is an alternative way to test in the future!

alyxadarr (commented):
Great job on this assignment. Your code looks clean. 
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔ line coverage was 76% and branch was 100%? not sure what "code coverage" entails. 
For this assignment, always use Assert.Equal<T>()  (the generic version) ❌  Mark did wave this rule on Microsoft Teams, so you should be good :)

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
connerhanson: I tried to implement this but quickly ran into problems.  When I referenced the GetJoke() method after creating a moq JokeService it was not able to locate the method.  After a little digging I found that when using moq on a class method the method should be marked as virtual.  When doing that I ended up running into a few problems.  Because of that I don't think I'll implement this test.  I also have a test that does exactly that using the interfaces and Jester class within JesterTests.  For the most part this test class/file was more meant to test that I am receiving something from the JokeService class.  Though I do appreciate  the feedback!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
connerhanson: Done, Thanks!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
connerhanson: Since this test is dependent on what you specify to return wouldn't you already know the amount of times it skipped since you are the one creating the return values?  I see the use for it in testing Output, but as long as it doesn't return/output a Chuck Norris joke when testing you should be good.  I like the suggestion, but it seems like more of a feature to see frequency of jokes that contain Chuck Norris from the http source.

alyxadarr: True Point!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,67 @@
...
+        string testJoke = "test joke";
+        moqJokeService.SetupSequence(x => x.GetJoke()).Returns(chuckJoke).Returns(testJoke);
+        Jester jesterTest = new(moqJokeService.Object, moqJokeOutput.Object);
+
+        // Act 
+        jesterTest.TellJoke();
+
+        // Assert with moq
+        // Should skip chuckJoke and equal testJoke
+        moqJokeOutput.Verify(x => x.OutputJoke(testJoke));
```
BenjaminMichaelis, line 64: This actually doesn't check that chuckJoke wasn't run.

You want something like:
```suggestion
        moqJokeOutput.Verify(x => x.OutputJoke(It.Any<string>()), Times.Once);

```
This makes sure for any input to OutputJoke that it was only called one. You just made sure that OutputJoke was called at all with the output of testJoke.

CanHazFunny/CanHazFunny.Tests/JokeOutputTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        // Arrange
+        string testJoke = "test joke";
+        JokeOutput jokeOutput = new(); 
+        StringWriter outputJoke = new();  //Needs to be string writer since we need to take in System.IO.TextWriter 
+        Console.SetOut(outputJoke);
+        jokeOutput.OutputJoke(testJoke);
+
+
+        // Act & Assert
+        Assert.Equal(testJoke + "\n" , outputJoke.ToString());
```
BenjaminMichaelis, line 22:
```suggestion
        // Act & Assert
        Assert.Equal(testJoke + Environment.NewLine , outputJoke.ToString());

```
This works on Linux but not on windows

PR #111, opened 2024-02-02
Assignment4 - Cole Stewart and Nathan Marsee

colestewart12:
@NathanMarsee @colestewart12


StrangeRanger:
@colestewart12 Just a reminder, don't forget to add your name and your partner's name to the title, the, @ each other in the message box with "Draft"

Aramirez61:
I'll review this.

Logger/FullName.cs
```diff
@@ -0,0 +1,51 @@
...
+            LastName = names[1];
+        }
+        else if (names.Length == 3)
+        {
+            FirstName = names[0];
+            MiddleName = names[1];
+            LastName = names[2];
+        }
+        else
+        {
```
Aramirez61, line 35: Consider specifying here that first or last name cannot be null so the user knows the error, and perhaps throw a null argument exception.

Logger/Student.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Logger;
+
+public record Student : IEntity
+{
+    private FullName _name;
+    public int Sid { get; init; }
+    public Guid Id { get; init; }
+    public string Name
```
Aramirez61, line 8: Consider using a separate Person record class that uses a calculated property for name so common code isnt written twice between student and employee

Logger.Tests/StudentTests.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
```
Aramirez61, line 4: Ensure that you are testing the equality behavior on all entities works as expected

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,198 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class StorageTests
```
Aramirez61, line 5: good job on thoroughly testing the behavior of Storage methods on all entities

Aramirez61 (commented):
Tasks
Provide a comment on the full name record on why or why not the type is immutable. ❌ (add this to the FullName record)
Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌
(Need to add these comments in Student, Employee, and Book entities)
You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
Test that the equality behavior on these entities behaves as expected. ❌
(Need these in all entity tests)
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
(Need these)
Fundamentals
Be sure you enable:
Extra Credit

Logger/Book.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace Logger;
+
+public record Book : IEntity
+{
```
Joshua-Pedersen, line 6: Don't forget to explain implicit or explicit implementation. ``provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly``

Logger/Student.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Logger;
+
+public record Student : IEntity
+{
```
Joshua-Pedersen, line 4: Comment needed, same as in ``Book``

Logger/Employee.cs
```diff
@@ -0,0 +1,30 @@
+﻿
+namespace Logger;
+
+//thoughts on this implementation of Employee/Student?
+public record Employee : IEntity
+{
```
Joshua-Pedersen, line 6: Comment needed, same as in ``Book``

Joshua-Pedersen (commented):
Issues found are quoted in the related code comments

## Tasks

- Define a full name record (first, last, middle) handling optional and null appropriately. ✔
  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Provide a comment on the full name record on why or why not the type is immutable. ❌
- Define an `IEntity` interface: ✔
  - Add an Id property of type `Guid` that is an init-only setter. ✔
  - Add a Name property that is `string`. ✔
- Define an abstract base class that implements `IEntity` - appropriately choosing to implement the interface explicitly or implicitly. ✔
  - Do not implement the `Name` property in this abstract class. ✔
  - Do force any derived classes to provide an implementation for `Name`. ✔
- Define book, student, and employee records - all with entity capabilities. ✔
  - Choose a thoughtful approach for the value of `Name` for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ❌
  - Provide a comment on each interface **member** in each entity explaining why you implemented it implicitly or explicitly. ❌
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
  - Given the properties on the full name record, you should consider which entities it makes sense to use it. ✔
  - Test that the equality behavior on these entities behaves as expected. ❌
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
- Unit test the Storage class using various entities to ensure that there are no bugs.
- Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.)
  - Pay special attention to the CreateLogger methods throughout.

## Fundamentals

- Be sure you enable:
  - Enable Nullability for all projects  ✔
  - Ensure that you turn on code analysis for all projects(`EnableNETAnalyzers`) ✔
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
  - Ensure that you turn on Nullability (`Nullable`) ✔
- **Ensure there are no errors or warnings (including code analysis warnings)** ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔

## Extra Credit

- Refactor your project files to put common properties and/or items into a single file. ❌
- Convert `ILogger.CreateLogger` to a generic method. ❌
  - Returns the logger classes type that is to be created. ❌
  - Has a type parameter for the configuration type ❌

Logger/FullName.cs
NathanMarsee: Good call, will do.

Logger.Tests/StudentTests.cs
NathanMarsee: Will do

Logger/FullName.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Text;
+
+namespace Logger;
+
+// value type, since it is more helpful to check equality of a name's data (the actual name) than its reference (the location in memory)
```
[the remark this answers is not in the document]
NathanMarsee: Thanks, will do

Logger/Student.cs
NathanMarsee: I planned on doing this, but thought that since they only have three lines of code shared between them, the separate Person record class would have been longer. 
You're right though, it is one of the requirements and probably right from a design perspective anyway.

Logger/Book.cs
NathanMarsee: You're right, we'll fix that

Logger/Student.cs
```diff
@@ -0,0 +1,33 @@
...
+    private FullName _name;
```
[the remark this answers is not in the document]
NathanMarsee: I could very well be wrong, but I thought my implementation here qualified as a calculated property, not a backing field. I'll look into doing it the way you suggest here, but I'm having trouble imagining how I'll store the name of one of these objects for the purposes of a calculated property without using some kind of field.

Logger/Book.cs
```diff
@@ -0,0 +1,35 @@
...
+    private FullName _author;
```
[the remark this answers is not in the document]
NathanMarsee: See my comment on Student.cs

Logger/Employee.cs
```diff
@@ -0,0 +1,30 @@
...
+    private FullName _name;
```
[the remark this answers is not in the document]
NathanMarsee: See my comment on Student.cs

Logger/Student.cs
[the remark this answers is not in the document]
BenjaminMichaelis: This is a backing field to a property, not a calculated property

Logger.Tests/IEntityRecordTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        Guid guid = Guid.NewGuid();
+        Book book1 = new(name, "test author1", 1234567890, guid);
+        Book book2 = new(name, "test author2", 1234567890, guid);
+
+        //Assert
+        Assert.False(book1 == book2);
+    }
+
+    [Fact]
+    public void Book_DifferentISBN_NotEqual()
```
BenjaminMichaelis, line 124: What if the guid's are different?

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,198 @@
...
+{
+    private sealed class TestEntity : IEntity
+    {
+        public Guid Id { get; init; }
+        public required string Name { get; set; }
+    }
+
+    [Fact]
+    public void AddandContains_BaseEntity_SuccessfullyAdded()
```
BenjaminMichaelis, line 14:
```suggestion
    public void Add_BaseEntity_SuccessfullyAdded()
```
1. Naming should be PascalCase
2. First part of the name should be the thing (generally method) under test (SUT)

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        //Arrange
+        var storage = new Storage();
+        var book = new Book("Book Name", "Author Name", 1234567890);
+        storage.Add(book);
+
+        //Act
+        storage.Remove(book);
+
+        //Assert
+        Assert.True(!storage.Contains(book));
```
BenjaminMichaelis, line 99:
```suggestion
        Assert.False(storage.Contains(book));
```
Simplify with False

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        var name = new FullName(first, last);
+        var Eid = 1234567890;
+        var employee = new Employee(name, Eid);
+        storage.Add(employee);
+
+        //Act
+        storage.Remove(employee);
+
+        //Assert
+        Assert.True(!storage.Contains(employee));
```
BenjaminMichaelis, line 147:
```suggestion
        Assert.False(storage.Contains(employee));
```
Simplify

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,198 @@
...
+        var name = new FullName(first, last);
+        var student = new Student(name, Sid);
+
+        storage.Add(student);
+
+        //Act
+        storage.Remove(student);
+
+        //Assert
+        Assert.True(!storage.Contains(student));
```
BenjaminMichaelis, line 196:
```suggestion
        Assert.False(storage.Contains(student));
```

Logger.Tests/IEntityRecordTests.cs
```diff
@@ -0,0 +1,137 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class IEntityRecordTests
+{
+    [Fact]
+    public void Student_SameStudents_AreEqual()
```
BenjaminMichaelis, line 8: These different types ideally should be in different files (StudentTests, EmployeeTests,BookTests,etc). You have no such type of "IEntityRecord"

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+
+namespace Logger;
+
+public record Book : IEntity
+{
+    // implemented implicitly since Name is a frequently-accessed member
+    public string Name { get; set; }
+
+    //TODO: explain why this is implemeted implicitly
```
BenjaminMichaelis, line 10: Seems like todo didn't get replaced

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+    public Guid Id { get; init; }
+
+    private FullName _author;
+        public string Author
+    {
+        get => _author.ToString();
+        set => _author = new(value);
+    }
```
BenjaminMichaelis, line 18:
```suggestion
      public FullName Author {get; set;}
```
A auto-generated property seems simpler here. Also, what is the purpose of this, is it used anywhere?

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+    public int Isbn { get; init; }
+
+    public Book(string name, string author, int isbn, Guid id)
+    {
+        Id = id;
+        Name = name;
+        _author = new(author);
```
BenjaminMichaelis, line 25: Remember, we access fields through their properties
```suggestion
        Author = new(author);
```

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+        Isbn = isbn;
+    }
+
+    public Book(string name, FullName author, int isbn, Guid id)
+    {
+        Id = id;
+        Name = name;
+        _author = author;
```
BenjaminMichaelis, line 33:
```suggestion
        Author = author;
```

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+        Name = name;
+        _author = new(author);
+        Isbn = isbn;
+    }
+
+    public Book(string name, FullName author, int isbn)
+    {
+        Id = Guid.NewGuid();
+        Name = name;
+        _author = author;
```
BenjaminMichaelis, line 49:
```suggestion
        Author = new(author);
        Isbn = isbn;
    }

    public Book(string name, FullName author, int isbn)
    {
        Id = Guid.NewGuid();
        Name = name;
        Author = author;
```

Logger/Employee.cs
```diff
@@ -0,0 +1,44 @@
+﻿
+namespace Logger;
+
+public record Employee : IEntity
+{
+    private FullName _name;
+    public int Eid { get; set; }
```
BenjaminMichaelis, line 7:
```suggestion
    public int EmployeeId { get; set; }
```
`Eid` seems like a not common shorthand, the full name is more clear

Logger/Employee.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    //TODO: explain why this is implemeted implicitly
+    public Guid Id { get; init; }
+
+    // implemented implicitly since Name is a frequently-accessed
+    public string Name
+    {
+        get => _name.ToString();
+        set => _name = new(value);
+    }
```
BenjaminMichaelis, line 17:
```suggestion
    public string Name { get => _name.FirstName + " " + _name.LastName; }
```
Use a calculated property instead. This would be even easier with a `Person` object as some reviewers have mentioned

Logger.Tests/BaseEntityTests.cs
```diff
@@ -0,0 +1,29 @@
...
+    public void BaseEntity_SetName_SuccessfulSet()
+    {
+        //Arrange
+        var testEntity = new TestEntity("Test Name");
+
+        //Act
+        ((IEntity)testEntity).Name = "New Name";
+
+        //Assert
+        Assert.Equal("New Name", ((IEntity)testEntity).Name);
```
BenjaminMichaelis, line 27:
```suggestion
        //Arrange/Act
        TestEntity testEntity = new("Test Name")
        {
            Name = "New Name"
        };
        
        //Assert
        Assert.Equal("New Name", testEntity.Name);
```
this can be simplified

Logger/FullName.cs
```diff
@@ -0,0 +1,67 @@
...
+// not immutable, as someone's name can change
+public record class FullName
```
BenjaminMichaelis, line 7:
```suggestion
public record struct FullName
```
 I think a record struct could of been more appropriate here

BenjaminMichaelis, line 7: This class probably could of been simplified:

```C#
public record struct FullName(string FirstName, string LastName, string? MiddleName = null)
{
    public string FirstName { get; } = FirstName??throw new ArgumentNullException(nameof(FirstName));
    public string LastName { get; } = LastName??throw new ArgumentNullException(nameof(LastName));
    public string? MiddleName { get; } = MiddleName;
}
```

Logger/FullName.cs
```diff
@@ -0,0 +1,67 @@
...
+            LastName = names[2];
+        }
+        else if (names.Length >= 2 && (names[0] is null || names[1] is null))
+        {
+            throw new ArgumentException("first name and last name can't be null");
+        }
+        else
+        {
+            throw new ArgumentException("invalid name");
+        }
```
BenjaminMichaelis, line 53: This doesn't seem like sound logic. what if a name has two first names? Or two last names? The error handling doesn't seem like it can be sufficient to make assumptions for all cases and other constructors you have would be better to be used.

Logger/Student.cs
```diff
@@ -0,0 +1,49 @@
...
+    public int Sid { get; init; }
```
BenjaminMichaelis, line 6:
```suggestion
    public int StudentId { get; init; }
```
This is more clear than `Sid`. We don't have to worry about performance or something when it's shorter, so we should opt for readability. If I'm a calling user, `Sid` could be confusing

Logger/Student.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        get
+        {
+            return _name.ToString();
+        }
+        set
+        {
+            _name = new(value);
+        }
+    }
```
BenjaminMichaelis, line 22: Again, this would be easier with a Person class

Logger/Student.cs
```diff
@@ -0,0 +1,49 @@
...
+    public Student(FullName name, int sid, Guid id)
+    {
+        _name = name;
```
BenjaminMichaelis, line 25:
```suggestion
        Name = name;
```
fields should be accessed through Properties. Comment applies to all places in file

Logger/Storage.cs
```diff
@@ -20,10 +20,8 @@ public bool Contains(IEntity item)
...
     public IEntity? Get(Guid expectedGuid)
     {
-        return Entities.FirstOrDefault(entity => 
-        {
-            dynamic dynamicEntity = entity;
-            return dynamicEntity.Id == expectedGuid;
-        });
+        return Entities
+        .OfType<IEntity>()
+        .FirstOrDefault(entity => entity.Id == expectedGuid);
```
BenjaminMichaelis, line 25:
```suggestion
        return Entities.FirstOrDefault(entity => entity.Id == expectedGuid);
```
This can be simplified

PR #117, opened 2024-02-05
Assignment4 - Johanne McClenahan and Alexa Darrington

alyxadarr:
@JohanneJayde Johanne's github 
@alyxadarr Alexa's github


Purplepigs0129:
Reviewing!

Logger/ILogger.cs
```diff
@@ -1,12 +1,16 @@
...
 
     // While interesting, this is probably better implemented using a factory class.
     // because you can't have static abstract members on classes
     // and you can't have covariant return types on interface members. :(
-    static abstract ILogger CreateLogger(in ILoggerConfiguration configuration);
+    //NOTE: This is the EC attempt but it is not correct. We have left it here to see if
+    //any other groups has done their own implmentation that works.
+#pragma warning disable CA1000
+    static abstract T CreateLogger<T2>(in T2 configuration) where T2 : ILoggerConfiguration;
+#pragma warning restore CA1000
```
Purplepigs0129, line 15: I see the EC attempt I think it is converted to a generic method correctly but I don't think it is returning the proper type or has the correct type parameter.

Logger/Book.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Logger;
+
+public record class Book(string Title, long? Isbn) : IEntity
+{
+    public string Title { get; init; } = Title ?? throw new ArgumentNullException(nameof(Title));
```
Purplepigs0129, line 5: Interesting additions here, we only hit the name so might have to add some more to ours. Good Work!

Logger/Employee.cs
```diff
@@ -0,0 +1,15 @@
+﻿
+namespace Logger;
+    public record class Employee(string FirstName, string LastName, int? EmployeeID) : Person(FirstName, LastName), IEntity
+    {
+
+        public int? EmployeeID { get; } = EmployeeID ?? throw new ArgumentNullException(nameof(EmployeeID));
```
Purplepigs0129, line 6: Nice Job using the ID here, still working on ours:)

Logger/FileLogger.cs
```diff
@@ -1,25 +1,31 @@
 ﻿namespace Logger;
 
-public class FileLogger : BaseLogger, ILogger
+//TODO: The in part of the CreateLogger method could be removed. I think this is because that means the loggerConfiguration is
+//contravarient. So it could specify less derrived class of ILogConfiguration except ILogConfiguration doesn't have any.
+//If you supply a ILogConfiguration type then it would break since it doesn't have the proper propreties to create a
+//FileLogger
+public class FileLogger : BaseLogger, ILogger<FileLogger>
 {
     private FileInfo File { get; }
```
Purplepigs0129, line 9: I'm curious as to why changes were needed here unless this was for the EC or was just needed for the way you guys are doing this.

Logger/Person.cs
```diff
@@ -0,0 +1,11 @@
+﻿
+namespace Logger;
+
```
Purplepigs0129, line 3: Awesome work here we did something similar

Purplepigs0129 (commented):
Tasks
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
Fundamentals
Be sure you enable:
Extra Credit
Convert ILogger.CreateLogger to a generic method. ✔

All of the common properties aren't in a single file, didn't know how to comment on this one except for here. I'm not quite too sure how to judge simplicity but from what I am seeing and how you guys did it, it seems as though you guys actually wrote some simple code. Also nice attempt on the extra credit I think the last fixes will become more apparent after the next class. Great Work! :)

Logger/Person.cs
alyxadarr: thank you!

Logger/ILogger.cs
JohanneJayde: I think it is. Since it's T, you can specify which type you want it to be. We added a type constraint so that T would only be classes that derive from BaseLogger. So if you put T as FileLogger then the CreateLogger method would return a FileLogger object.

Logger/Book.cs
JohanneJayde: Thanks! We initially didn't have ISBN, but after Benjamin and Mark talked about having a secondary human-identifiable ID, we added in ISBN so that users are able to see a unique ID for books and not just the id

Logger/Employee.cs
JohanneJayde: Thank you. I honestly think that this could be improved because we have to prefix the Employee Id with its name which I think is kind of clunky. I might try and figure out a better way of making Id explicit. I know you guys also ran into the same issue as us.

Logger/FileLogger.cs
JohanneJayde: Yes, this was for EC. ILogger became generic so we had to change this line.

JohanneJayde:
We attempted EC on generics. This was the best we could do. I am very interested in seeing the best way to implement it and what things we could change to improve our version.

Logger/ILogger.cs
JohanneJayde: Okay so after a lot of thinking and working. This was not correct at all. Having static abstract methods in a generic interface is a code analyzer issue. We ended up doing a different way that involved using a factory method and class.

Logger/Student.cs
```diff
@@ -0,0 +1,18 @@
+﻿
+namespace Logger;
+
+public record class Student(string FirstName, string LastName, int? StudentID) : Person(FirstName, LastName), IEntity
+{
+
+    public int? StudentID { get; init; } = StudentID ?? throw new ArgumentNullException(nameof(StudentID));
+
+    //Name member of IEntity should explicit becuase it has collision with Employee's name
+    string IEntity.Name {get => $"EntityType: {nameof(Student)}, EntityID: {Id}, Name: {Name}, StudentID: {StudentID}"; }
```
BenjaminMichaelis, line 10: I would argue that Name can just be like firstname + lastname or similar

Logger.Tests/EmployeeTests.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+    public class EmployeeTests
+    {
+    #pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
```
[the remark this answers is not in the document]
alyxadarr: good to know!

Logger.Tests/EmployeeTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        var employee1 = new Employee("Thomas", "Young", 11123) { 
+            Id = entityID,
+        };
+        var employee2 = new Employee("Thomas", "Young", 11123)
+        {
+            Id = entityID,
+        }
+;
+        Assert.True(employee1.Equals(employee2));
+        Assert.True(employee1 == employee2);
```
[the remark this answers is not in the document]
alyxadarr: noted!

PR #118, opened 2024-02-05
Assignment4 - Paulson Ray Thompson & Trevor Stewart

Purplepigs0129:
Trevor's Git - @Tvor369


JohanneJayde:
I will review this!

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record class Person : BaseEntity
+{
+    //Implemented Name implcitily since FullName is never saved as a property itself so it cant be accessed, so no conflicts
+    public override string Name { get; set; }
```
JohanneJayde, line 12: I think your approach of having the Name just be the student's name, employee's name or title is not the best way of doing it. It doesn't differentiate them enough. What we did was add the type of entity to the Name field so that you could see that was along with the name, title, etc. It adds more details. Also, consider adding the Id field but I don't know if we will keep it in there in our implementation.

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
...
+
+    public Person(FullName name)
+    {
+        Name = name.ToString();
+    }
```
JohanneJayde, line 17: This doesn't check if FullName is null.
```suggestion
    public Person(FullName name)
    {
ArgumentException.ThrowIfNullOrWhiteSpace(name);        Name = name.ToString();
    }
```
This is if you decide to stay with this constructor. If you do a primary constructor, you can do a null coalescing operator for null-checking for name

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record class Person : BaseEntity
```
JohanneJayde, line 9: Since Person is not abstract, you may want to make tests for it like checking for null values and equity since it is now also an Entity, or consider making it abstract.

Logger/Book.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+//Implimented implicitly since Name isn't causing any collisions, and if we ever need an Author name we would
+// probably add an Author property of Entity type person which would get the Author name through that.
+
+public record class Book : BaseEntity
```
JohanneJayde, line 14: nit: I would also put why Id is implicit in each Entity class instead of just having the explanation in BaseEnity for clairty.

JohanneJayde (commented):
Provide a comment on the full name record on why or why not the type is immutable. ❌ Explanation was why it should be mutable instead of if the implemented FullName record was mutable or immutable
Define an abstract base class that implements IEntity - appropriately choosing to implement the interface explicitly or implicitly. ❌ Missing comment on why Name should be implemented implicitly.
Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ❌ Name field does have a backing field in all of the Entities.
Test that the equality behavior on these entities behaves as expected. ❌ Missing tests for checking when different Entities are equal to each other, not just not equal.
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ✔
Unit test the Storage class using various entities to ensure that there are no bugs. ✔
Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.). Pay special attention to the CreateLogger methods throughout. ✔ This is done but there are other places in Logger classes that one could find issues with.
Extra Credit

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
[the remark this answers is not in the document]
Purplepigs0129: You would be correct will fix this

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,96 @@
...
+using Xunit;
+namespace Logger.Tests;
+
+public class StorageTests
+{
+    //Student
```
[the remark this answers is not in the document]
Purplepigs0129: Ya I had this in my head but got confused due to the MSUnit mix up thought it was public Constructor() but its that thank you

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Logger;
+
+public abstract record class BaseEntity : IEntity
+{
+    //I think we should implement Id explicitly since Entities like Student may later have a Student Id which could be confusing.
+    //However our project then runs into the problem where we cannot get a derived entity's Id since it has to be marked private.
+    // This caused problems in StorageTests when testing the Get method...
+    public Guid Id { get; init; } = Guid.NewGuid();
+    //Guid IEntity.Id { get; init; } = Guid.NewGuid();
+    public abstract string Name { get; set; }
```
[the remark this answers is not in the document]
Purplepigs0129: WIll add this

Logger/Book.cs
Purplepigs0129: will add in whilst adding in ID

Logger/FullName.cs
```diff
@@ -0,0 +1,45 @@
...
+    {   //we definetly talked about a more effcient way in class but this works at least...
+        if(MiddleName != null)
+        {
+            return FirstName + " " + MiddleName + " " + LastName;
+        }
+        else
+        {
+            return FirstName + " " + LastName;
+        }
+    }
```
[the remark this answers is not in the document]
Purplepigs0129: yep this is more effiecient thank you

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    [Fact]
+    public void Student_InitializeStudentWithMiddleName_Success()
+    {
+        FullName fullName = new("First", "Middle", "Last");
+        Student student = new(fullName);
+        Assert.Equal("First Middle Last", student.Name);
+    }
+
```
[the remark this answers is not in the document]
Purplepigs0129: I thought I did this already must've gotten overwritten ina commit thanks for the catch

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    [Fact]
+    public void Get_Book_ReturnsBooksID()
+    {
+        Storage storage = new();
+        Book book = new("Title");
+        storage.Add(book);
+        Assert.Equal(book, storage.Get(book.Id));
+    }
+}
```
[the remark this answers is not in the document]
Purplepigs0129: REally just testing for each method I can try to take some out and see if the coverage stays but otherwise will probably leave in, also tested this way to make sure it is getting each entity back correctly

Logger/Book.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public override string Name { get; set; }
+
+    public Book(string title)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(title);
+        Name = title;   
+    }
+
+}
```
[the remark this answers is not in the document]
Purplepigs0129: Thanks for the fix glad we talked about this in class

Logger/FullName.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+
+namespace Logger;
+
+
+
+// I've changed this to a value type since it is more or less just a simple variable, the entities that
+// FullName will apply to should handle the uniqueness of the object.
+
+// I think FullName should be a mutable type? Since people can change their name.
```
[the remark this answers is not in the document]
Purplepigs0129: Will update comment to match correctly with the assignment details so its less of a question and more of a direct statement

Logger/IEntity.cs
```diff
@@ -1,7 +1,7 @@
-﻿namespace Logger;
+namespace Logger;
 public interface IEntity
 {
-
         // Place members here.
-
+        public Guid Id { get;  init; }
+        public string Name { get; set; }
```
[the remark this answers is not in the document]
Purplepigs0129: Yes was set this way before class on Tuesday didn't get a chance to fix in afterwards will fix

Logger/Person.cs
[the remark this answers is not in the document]
Purplepigs0129: Just kind of made sense to do it this way that way Emplyoee and Student who are deriving from this have access to the entity as well

Purplepigs0129: Will do just ran out of time on tuesday

Logger/Person.cs
Purplepigs0129: Will think about ID and yes we shall think of ways to differentiate the name field so that it stands out more but might not depending

Logger/Person.cs
Purplepigs0129: nice catch will fix

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    //Implemented Name implcitily since FullName is never saved as a property itself so it cant be accessed, so no conflicts
+    public override string Name { get; set; }
+
+    public Person(FullName name)
+    {
+        Name = name.ToString();
+    }
+}
```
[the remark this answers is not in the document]
Purplepigs0129: thanks just ran out of time on tuesday after class

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Logger;
+
+public record class Person : BaseEntity
+{
+    //Implemented Name implcitily since FullName is never saved as a property itself so it cant be accessed, so no conflicts
+    public override string Name { get; set; }
+
+    public Person(FullName name)
+    {
+ArgumentException.ThrowIfNullOrWhiteSpace(name);        Name = name.ToString();
```
Tvor369, line 16:
```suggestion
     Name = name.ToString();
```
The parts in Name will actually be check within FullName indiviually so this shouldnt be needed. 
Plus the name passed in through the constructor is not just a string it is a FullName struct so it will break with a this Exception check.

Logger/FullName.cs
```diff
@@ -0,0 +1,33 @@
...
+        //middle can be null as defualt since some people dont have one, thus it should be the only nullable
+
+        FirstName = firstName;
+        MiddleName = middleName;
+        LastName = lastName;
+    }
+
+    public override string ToString()
+    {   //we definetly talked about a more effcient way in class but this works at least...
+    return $"{FirstName {(string.IsNullOrWhiteSpace(MiddleName) ? String.Empty : " +  {MiddleName} {LastName}";
```
Tvor369:
```suggestion
    return $"{FirstName}{(string.IsNullOrWhiteSpace(MiddleName) ? String.Empty : " " +  MiddleName)} {LastName}";
```

Logger/FullName.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public readonly record struct FullName(string FirstName, string LastName, string? MiddleName = "")
+{
+    public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
+    public string MiddleName { get; } = MiddleName ?? "";
+    public string LastName { get; } = LastName ?? throw new ArgumentNullException(nameof(LastName));
+
+    public override string ToString()
+    {
+    return $"{FirstName}{(string.IsNullOrWhiteSpace(MiddleName) ? String.Empty : " +  MiddleName)} {LastName}")}";
```
Tvor369, line 18:
```suggestion
    return $"{FirstName}{(string.IsNullOrWhiteSpace(MiddleName) ? String.Empty : " "+  MiddleName)} {LastName}")}";
```

Logger/Person.cs
Tvor369: Ill make it abstract

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,219 @@
...
+        Student student1 = new(fullName)
+        {
+            ID = id,
+        };
+
+        Assert.Equal(student.ID, student1.ID);
+    }
+
+    [Fact]
+    public void Student_InitialzeStudentsFirstNameWithNull_ThrowException()
```
BenjaminMichaelis, line 120:
```suggestion
    public void Student_InitializeStudentsFirstNameWithNull_ThrowException()
```

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        Assert.Throws<ArgumentNullException>(() => new Student(new(null!, "Chef", "Marcus")));
+    }
+
+    [Fact]
+    public void Student_InitialzeStudentsLastNameWithNull_ThrowException()
```
BenjaminMichaelis, line 126:
```suggestion
    public void Student_InitializeStudentsLastNameWithNull_ThrowException()
```

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        Assert.Throws<ArgumentNullException>(() => new Student(new("Chef", null!, "Terry")));
+    }
+
+    [Fact]
+    public void Student_InitialzeStudentsFirstNameWithBlank_ThrowException()
```
BenjaminMichaelis, line 132:
```suggestion
    public void Student_InitializeStudentsFirstNameWithBlank_ThrowException()
```

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        Assert.Throws<ArgumentException>(() => new Student(new(" ", "Michael", "Berzatto")));
+    }
+
+    [Fact]
+    public void Student_InitialzeStudentsLastNameWithBlank_ThrowException()
```
BenjaminMichaelis, line 138:
```suggestion
    public void Student_InitializeStudentsLastNameWithBlank_ThrowException()
```

Logger.Tests/EntityTests.cs
BenjaminMichaelis, line 1: Student, Book, Employee, etc tests should all be in their own file. One set of tests for every model per file

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        Storage = new();
+    }
+
+    //Student
+    [Fact]
+    public void Add_Student_ContainsStudent()
+    {
+        Student student = new(new FullName("Dean", "Winchester", null));
+        Storage.Add(student);
+        Assert.True(Storage.Contains(student));
```
BenjaminMichaelis, line 18: `Assert.Contains()` is probably more clear and better here

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    [Fact]
+    public void Remove_Student_DoesNotContainStudent()
+    {
+        Student student = new(new FullName("Samuel", "Winchester", "Mike"));
+        Storage.Add(student);
+        Storage.Remove(student);
+        Assert.False(Storage.Contains(student));
```
BenjaminMichaelis, line 27: `Assert.DoesNotContain()` is probably more clear and better here

Logger/Book.cs
```diff
@@ -0,0 +1,11 @@
+namespace Logger;
+
+public record class Book(string Title) : BaseEntity
+{
+    //Implimented implicitly since Name isn't causing any collisions, and if we ever need an Author name we would
+    // probably add an Author property of Entity type person which would get the Author name through that.
+    public override string Name { 
+    get=>$"{Title}";}
```
BenjaminMichaelis, line 8:
```suggestion
    public override string Name { get=> Title; }
```
this should simplify this to not need an interpolated string

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Logger;
+
+public abstract record class BaseEntity : IEntity
+{
+    //I think we should implement Id explicitly since Entities like Student may later have a Student Id which could be confusing.
+    //However our project then runs into the problem where we cannot get a derived entity's Id since it has to be marked private.
+    // This caused problems in StorageTests when testing the Get method...
+    public Guid ID { get; init; } = Guid.NewGuid();
```
BenjaminMichaelis, line 8: Why is an init; needed here?

Logger/FullName.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
+    public string MiddleName { get; } = MiddleName ?? "";
+    public string LastName { get; } = LastName ?? throw new ArgumentNullException(nameof(LastName));
+
+    public override string ToString()
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(FirstName);
+        ArgumentException.ThrowIfNullOrWhiteSpace(LastName);
+        return $"{FirstName}{(string.IsNullOrWhiteSpace(MiddleName) ? String.Empty : $" {MiddleName}")} {LastName}";
```
BenjaminMichaelis, line 20:
```suggestion
        return $"{FirstName}{(string.IsNullOrWhiteSpace(MiddleName) ? string.Empty : $" {MiddleName}")} {LastName}";
```
`string` over `String`

Logger/IEntity.cs
```diff
@@ -1,7 +1,7 @@
-﻿namespace Logger;
+namespace Logger;
 public interface IEntity
 {
-
         // Place members here.
-
+        public Guid ID { get;  init; }
+        public abstract string Name { get; }
```
BenjaminMichaelis, line 6: Interfaces should not have accessibility modifiers
```suggestion
        Guid ID { get;  init; }
        abstract string Name { get; }
```

PR #121, opened 2024-02-06
Assignment4 - Joshua Merrell and Trevor Felchlin 

Tfelchlin13:
github accounts: @tfelchlin13 and @joshmerrell


Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,66 @@
...
+public sealed class TestEntity : IEntity
+{
+    public Guid Id { get; init; } = Guid.NewGuid();
+    public string Name { get; set; } = "";
+}
+
+public class StorageTests
+{
+    [Fact]
+    public void StorageAdd_ContainsAdded()
```
SaaWest, line 19: Why did you instantiate a a TestEntity object? Why not a Record object type like Student, book or employee and store that?

Logger/FullName.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public void Deconstruct(out string First, out string Last, out string? Middle)
+    {
+        First = this.First;
+        Last = this.Last;
+        Middle = this.Middle;
+    }
+    public override string ToString()
+    {
+        return $"{First} {((Middle!=null)?Middle + " ":"")}{Last}";
```
SaaWest, line 21: much more simple then the option shown in class. Nice!
```suggestion
        return $"{First} {((Middle!=null)?Middle + " ":"")}{Last}";
   can be written as
        return FirstName + $"{(MiddleName != null ? " " + MiddleName + " " : " ")}" + LastName;
```

Logger/Records.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public abstract record class BaseRecordEntity(string Name): IEntity
```
SaaWest, line 9: I think by convention the records are suppose to be in their own file but I'd check with Mark Michaelis

SaaWest (commented):
Tasks

        Do not implement the Name property in this abstract class. ❌
        You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
        Given the properties on the full name record, you should consider which entities it makes sense to use it. ❌
        Test that the equality behavior on these entities behaves as expected. ❌
        Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ❌
    Unit test the Storage class using various entities to ensure that there are no bugs.✔

Fundamentals

    Be sure you enable:
        Enable Nullability for all projects ❌✔ (has this in one csproj file but not the other)
        Ensure that you turn on Nullability (Nullable) ❌✔ (has this in one csproj file but not the other)

Extra Credit

NathanMarsee (approved):
Project Tasks:
  Do not implement the Name property in this abstract class. ✔ _(I'm not actually 100% sure on this, but it's what I did.)_
  You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌✔ _(has no common code at the moment)_
  Given the properties on the full name record, you should consider which entities it makes sense to use it. ❌
  Test that the equality behavior on these entities behaves as expected. ❌
  Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ✔
Unit test the Storage class using various entities to ensure that there are no bugs. ✔
Fundamentals:
Be sure you enable:
Enable Nullability for all projects ❌✔_(has this in one csproj file but not the other)_
Ensure that you turn on Nullability (Nullable) ❌✔ _(has this in one csproj file but not the other)_
Extra Credit:
  
Reviewer's Comments:
_Great work overall! just a few things to check up on and some possible restructuring of some of the records, but it's not far off._

Logger.Tests/RecordsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    public void TestEmployeeRecords(string name, decimal wage)
+    {
+        EmployeeRecord employee = new(name, wage);
+        Assert.Equal(name, employee.Name);
+        Assert.Equal(wage, employee.Wage);
+        Assert.True(employee.Equals(employee));
+        Assert.False(employee.Equals(new EmployeeRecord(name, wage)));
+    }
+}
+
```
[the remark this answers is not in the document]
Tfelchlin13: Thanks

Logger.Tests/StorageTests.cs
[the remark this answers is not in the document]
Tfelchlin13: Oh yea thanks for reminding me ill fix that

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+public abstract class BaseEntity : IEntity
+{
+    // Id is implemented implicitly, since BaseEntity only implements one interface
+    public Guid Id { get; init; } = Guid.NewGuid();
```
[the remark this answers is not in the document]
Tfelchlin13: In this particular scenario, implicit implementation is chosen to keep things simple and easy to understand. With just one interface being implemented, there's no confusion about where the Id property comes from. but i understand why you would do it explicitly

Logger/FullName.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+// FullName is a reference type because we do not want the data to be immutable, as a person's
+// name can change (e.g., when they get married). As a record class, the data is not immutable.
+public record class FullName(string First, string Last, string? Middle = null)
```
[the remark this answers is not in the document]
Tfelchlin13: I would agree with that i swapped last and middle name.

Logger/FullName.cs
Tfelchlin13: thanks!

Logger/Records.cs
Tfelchlin13: i would agree it could be more clear if we had them separate i moved them each to their own file.

Logger/FullName.cs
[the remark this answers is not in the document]
Tfelchlin13: i forgot you cant do it that way because c# requires optional params to be after all the required params.

Logger.Tests/StorageTests.cs
joshmerrell: I did consider using one of those, but felt that using a clean slate as a stub made more sense.

Logger.Tests/FullNameTests.cs
```diff
@@ -0,0 +1,35 @@
...
+
+public class FullNameTests
+{
+    [Theory]
+    [InlineData("First", "Last", "Middle", "First Middle Last")]
+    [InlineData("First", "Last", null, "First Last")]
+    public void FullName_ToString_HandlesNullMiddle(string first,  string last, string? middle, string expects)
+    {
+        FullName name = new(first, last,middle);
+        Assert.Equal(expects, name.ToString());
```
BenjaminMichaelis, line 13:
```suggestion
    public void FullName_ToString_HandlesNullMiddle(string first,  string last, string? middle, string expected)
    {
        FullName name = new(first, last,middle);
        Assert.Equal(expected, name.ToString());
```
I think slightly better naming

Logger.Tests/Logger.Tests.csproj
```diff
@@ -7,6 +7,7 @@
     <LangVersion>12.0</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <EnforceCodeStyleInBuild>True</EnforceCodeStyleInBuild>
```
BenjaminMichaelis, line 10:
```suggestion
    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
nit: casing

Logger.Tests/RecordsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    [InlineData("Fear and Loathing in Las Vegas", "Hunter S. Thompson")]
+    [InlineData("2001: A Space Odysssey", "Arthur C. Clarke")]
+    public void TestBookRecords_BookNameAndAuthor_Successful(string bookName, string author)
+    {
+        BookRecord book = new(bookName, author);
+        BookRecord alsoBook = book;
+        BookRecord book2 = new(bookName, author);
+        Assert.Equal(bookName, book.Name);
+        Assert.Equal(author, book.Author);
+        Assert.True(book.Equals(alsoBook));
```
BenjaminMichaelis, line 24: Why not use `Assert.Equal` here?

Logger.Tests/RecordsTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    [Theory]
+    [InlineData("Inigo Montoya", "Fencing")]
+    [InlineData("Wesley", "Piracy")]
+    public void TestStudentRecords_NameAndMajor_Successful(string name, string major)
+    {
+        StudentRecord student = new(name, major);
+        Assert.Equal(name, student.Name);
+        Assert.Equal(major, student.Major);
+        Assert.True(student.Equals(student));
+        Assert.False(student.Equals(new StudentRecord(name, major)));
```
BenjaminMichaelis, line 37: Why not use `Assert.Equal` and `Assert.NotEqual`?

Logger/BaseEntity.cs
BenjaminMichaelis, line 11: Why is `init` needed here?

Logger/BaseRecordEntity.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Logger;
+
+public abstract record class BaseRecordEntity(string Name): IEntity
```
BenjaminMichaelis, line 3: What is the purpose of this that is different than `BaseEntity`? And also, record is not needed in the name since it is clear that it is of `record` type

Logger/StudentRecord.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Logger;
+//Implemented implicitly because since it only implements one interface, there was no need to explicitly implement the interface member
+public record class StudentRecord(string Name, string Major) : BaseRecordEntity(Name)
```
BenjaminMichaelis, line 3: Record is not needed in the name, `record` is clear by the type.

Logger/FullName.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public void Deconstruct(out string First, out string Last, out string? Middle)
+    {
+        First = this.First;
+        Last = this.Last;
+        Middle = this.Middle;
```
BenjaminMichaelis, line 17: Why do you need the `this` keyword? And also, parameters have camelCase naming

PR #123, opened 2024-02-06
Assignment4 Matthew adams and Rachael DeBoer

MattA41:
@MattA41 and @rkdeboer

### DIFF
--- a/Logger.Tests/EntityTests.cs
+++ b/Logger.Tests/EntityTests.cs
@@ -17,7 +17,7 @@ public class EntityTests
         FullName testName = new("Cotton", "Joe", "Eye");
 
 
-        Student student = new(testName);
+        Student student = new(testName, new Guid());
 
 
         Assert.Equal(nameTest, student.Name);
@@ -30,7 +30,7 @@ public class EntityTests
         FullName testName = new("Cotton", "Joe", "Eye");
 
 
-        Student employee = new(testName);
+        Student employee = new(testName, new Guid());
 
         Assert.Equal(nameTest, employee.Name);
 
@@ -39,8 +39,9 @@ public class EntityTests
     public void Book_NameMatches()
     {
         string titleTest = "Count your lucky stars";
-        Book book = new(titleTest);
+        Book book = new(titleTest, new Guid());
 
         Assert.Equal(titleTest, book.Title);
     }
+
 }
--- a/Logger.Tests/EntityTests.cs
+++ b/Logger.Tests/EntityTests.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Logger.Tests;
+
+public class EntityTests
+{
+    [Fact]
+    public void Student_GivenName_NameMatches()
+    {
+        
+        string nameTest = "Cotton Eye Joe";
+        FullName testName = new("Cotton", "Joe", "Eye");
+
+
+        Student student = new(testName);
+
+
+        Assert.Equal(nameTest, student.Name);
+
+    }
+    [Fact]
+    public void Employee_GivenName_NameMatches()
+    { 
+        string nameTest = "Cotton Eye Joe";
+        FullName testName = new("Cotton", "Joe", "Eye");
+
+
+        Student employee = new(testName);
+
+        Assert.Equal(nameTest, employee.Name);
+
+    }
+    [Fact]
+    public void Book_NameMatches()
+    {
+        string titleTest = "Count your lucky stars";
+        Book book = new(titleTest);
+
+        Assert.Equal(titleTest, book.Title);
+    }
+}
--- a/Logger.Tests/FullNameTests.cs
+++ b/Logger.Tests/FullNameTests.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Logger.Tests;
+
+public class FullNameTests
+{
+    [Fact]
+    public void FullName_NullMiddleName_StillValid()
+    {
+        FullName name = new("Cotton", "Joe");
+        string nameTest = "Cotton Joe";
+        Assert.Equal(nameTest, name.Name);
+    }
+    [Fact]
+    public void FullName_NonNullMiddleName_StillValid()
+    {
+        FullName name = new("Cotton", "Joe", "Eye");
+        string nameTest = "Cotton Eye Joe";
+        Assert.Equal(nameTest, name.Name);
+    }
+    [Fact] 
+    public void NoFirstName_ThrowsException()
+    {
+        
+        Assert.Throws<ArgumentNullException>(() => new FullName(null!, "Joe", "Eye"));
+    }
+    [Fact]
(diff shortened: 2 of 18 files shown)


montalcy:
Reviewing

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        FullName testName = new("Cotton", "Joe", "Eye");
+
+
+        Student employee = new(testName);
+
+        Assert.Equal(nameTest, employee.Name);
+
+    }
+    [Fact]
+    public void Book_NameMatches()
```
montalcy, line 39: I appreciate ya'll making a whole new test class, I put it all in one, but this seems a lot cleaner.

Logger.Tests/FullNameTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        Assert.Equal(nameTest, name.Name);
+    }
+    [Fact] 
+    public void NoFirstName_ThrowsException()
+    {
+        
+        Assert.Throws<ArgumentNullException>(() => new FullName(null!, "Joe", "Eye"));
+    }
+    [Fact]
+    public void NoLastName_ThrowsException()
```
montalcy, line 34: Maybe this could all be with EntityTest file since they all seem to be taking care of the name equality

montalcy, line 34: Good job catching the null exception!

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        storage.Add(student);
+
+        Assert.True(storage.Contains(student));
+        storage.Remove(student);
+        Assert.False(storage.Contains(student));
+    }
+    [Fact]
+
+    public void RemoveBook_BookRemoved()
```
montalcy, line 75: I forgot to test removed. Good catch!

Logger/Employee.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+public record Employee : Person
+{
+
+    public Employee(FullName name) : base(name) { }
```
montalcy, line 11: Do we not need to do the Guid?

Logger/FullName.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+    public class FullName(string firstName, string lastName, string? middleName = null)
+    {
+        public string FirstName { get; } = string.IsNullOrWhiteSpace(firstName) ? throw new ArgumentNullException(nameof(firstName)) : firstName;
+        public string LastName { get; } = string.IsNullOrWhiteSpace(lastName) ? throw new ArgumentNullException(nameof(lastName)) : lastName;
+        public string? MiddleName { get; } = middleName;
+        public string Name => (MiddleName == null ? $"{FirstName} {LastName}" : $"{FirstName} {MiddleName} {LastName}");
```
montalcy, line 13: Good efficient use of your code

Logger/FullName.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+    public class FullName(string firstName, string lastName, string? middleName = null)
```
montalcy, line 8: make sure to add your comments on nullability and explicit/implicit

Logger/Person.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+    public record class Person(FullName FullName) : BaseEntity
+    {
+        
+        public override string Name { get => FullName.Name; }
```
montalcy, line 12: I didnt know you could do the getter this way, nice.

Logger/Employee.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+public record Employee : Person
```
montalcy, line 8: Nice implementation of person record class to simplify the employee and student

Logger.Tests/EntityTests.cs
rkdeboer: Thank you! I think keeping them seprate is the key to organization

Logger.Tests/FullNameTests.cs
rkdeboer: While I understand this, I think the tests should remain seprate as one checks to see if the name in FullName is set and the other tests checks to see if the FullName is applied to entities. Good observation though

rkdeboer: Thank you!

Logger.Tests/StorageTests.cs
rkdeboer: Thank you!

Logger/FullName.cs
rkdeboer: Thanks!

Logger/FullName.cs
rkdeboer: Good catch I will add those comments!

Logger/Person.cs
rkdeboer: Thank you!

Logger/Employee.cs
rkdeboer: Thank you!

Logger/Employee.cs
rkdeboer: I missed that thank you!

Logger/Book.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+public record Book(string Title, Guid Id)  : BaseEntity
+{
+    public override string Name { get; } = Title;
+    private readonly Guid id = Id;
```
BenjaminMichaelis, line 12:
```
    private readonly Guid id = Id;
```
This is no the correct naming. Also, this should be a property, as discussed previously, this should not be a field

Logger/IEntity.cs
```diff
@@ -2,6 +2,7 @@
 public interface IEntity
 {
 
-        // Place members here.
+    Guid Id { get; init; }
```
BenjaminMichaelis, line 5: Why do you need an `init`?

Logger/Book.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Metadata.Ecma335;
```
BenjaminMichaelis, line 4: What is this used for?

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Logger.Tests;
+
+public class EntityTests
+{
+    [Fact]
+    public void Student_GivenName_NameMatches()
```
BenjaminMichaelis, line 13: All of these entities (Student, Employee, etc) should be in different files.

A model of tests should be in their own files

PR #124, opened 2024-02-06
Assignment4 - Omar Facundo and Manny Aguilera

omar5o9:
@omar5o9 and @MannyPoptart

### DIFF
--- a/Logger.Tests/StorageTests.cs
+++ b/Logger.Tests/StorageTests.cs
@@ -1,0 +1,44 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class StorageTests
+{
+
+    [Fact]
+    public void Employee_Is_Created_WithProperDepartmentAndID()
+    {
+        Employee testEm = new();
+        testEm.EmployeeId = 32;
+        testEm.Department = "IT";
+
+        Assert.Equal(32, testEm.EmployeeId);
+        Assert.Equal("IT", testEm.Department);
+
+    }
+
+    [Fact]
+    public void Book_IsCreated_WithProperTitleAuthorYearPublished()
+    {
+        Book testBook = new();
+        testBook.Title = "Test";
+        testBook.Author = "Santi";
+        testBook.YearPublished = 1989;
+
+        Assert.Equal("Test", testBook.Title);
+        Assert.Equal("Santi", testBook.Author);
+        Assert.Equal(1989, testBook.YearPublished);
+    }
+
+    [Fact]
+    public void Student_IsCreated_WithProperId()
+    {
+        Student testStudent = new();
+        testStudent.StudentId = 31;
+
+        Assert.Equal(31, testStudent.StudentId);
+    }
+
+    
+}
+
--- a/Logger/BaseClass.cs
+++ b/Logger/BaseClass.cs
@@ -1,0 +1,11 @@
+﻿namespace Logger;
+
+public abstract class BaseClass : IEntity
+{
+
+    public abstract Guid Id { get; init; }
+
+    public abstract string Name { set; get; }
+
+}
+
--- a/Logger/BaseClass.cs
+++ b/Logger/BaseClass.cs
@@ -1,9 +1,9 @@
 ﻿namespace Logger;
 
-public abstract class BaseClass : IEntity
+public abstract record BaseClass : IEntity
 {
 
-    public abstract Guid Id { get; init; }
+    public Guid Id { get; init; }
 
     public abstract string Name { set; get; }
 
--- a/Logger/Book.cs
+++ b/Logger/Book.cs
@@ -1,0 +1,15 @@
+﻿
+namespace Logger
+{
+    public class Book : BaseClass
+    {
+        public string Title { get; set; }
+        public string Author { get; set; }
+        public int YearPublished { get; set; }
+
+        // Calculated property for the full name of the book (Title + Author)
+        public string FullName => $"{Title} by {Author}";
+
+        // Comment: The calculated property makes sense here as the full name is a combination of Title and Author.
+    }
+}
--- a/Logger/Book.cs
+++ b/Logger/Book.cs
@@ -1,15 +1,17 @@
 ﻿
-namespace Logger
+namespace Logger;
+
+public record Book : BaseClass
 {
-    public class Book : BaseClass
-    {
-        public string Title { get; set; }
-        public string Author { get; set; }
-        public int YearPublished { get; set; }
+    public string Title { get; set; }
+    public string Author { get; set; }
+    public int YearPublished { get; set; }
 
-        // Calculated property for the full name of the book (Title + Author)
-        public string FullName => $"{Title} by {Author}";
+    // Calculated property for the full name of the book (Title + Author)
+    public string FullName => $"{Title} by {Author}";
 
-        // Comment: The calculated property makes sense here as the full name is a combination of Title and Author.
-    }
+    public override string Name { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+    // The calculated property makes sense here as the full name is a combination of Title and Author.
 }
+
--- a/Logger/Employee.cs
+++ b/Logger/Employee.cs
@@ -1,11 +1,12 @@
 ﻿
-namespace Logger
-{
-    public class Employee : Person
-    {
-        public int EmployeeId { get; set; }
-        public string Department { get; set; }
+namespace Logger;
 
-        // Comment: Inheriting from Person to reuse the common code for FirstName, LastName, and FullName.
-    }
+public record Employee : Person
+{
+    public int EmployeeId { get; set; }
+    public string Department { get; set; }
+    public override string Name { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+    // Inheriting from Person to reuse the common code for FirstName, LastName, and FullName.
 }
+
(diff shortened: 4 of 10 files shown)


EthanAGi:
I'll review!

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        Student testStudent = new()
+        {
+            StudentId = 31
+        };
+
+        Assert.Equal(31, testStudent.StudentId);
+    }
+
+
```
EthanAGi, line 48: I'd recommend adding more tests, maybe testing for null and empty variables?

Logger/BaseClass.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Logger;
+
+public abstract record BaseClass : IEntity
+{
+
+    public Guid Id { get; init; }
+
+    public abstract string Name { set; get; }
```
EthanAGi, line 8: This is what I'm talking about, I think since you have a setter here the name may be implemented within the interface. I'm unsure on that though.

Logger/FullName.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+
+    // We decided to handle the full name storage this way to handle middle names being in the name or not
+    // We are allowing the name to be mutable because names can be changed
+    private string fullName;
+
+    public FullName(string fName, string mName, string lName)
+    {
+
+        fullName = $"{fName} {mName?.Trim()} {lName}";
```
EthanAGi, line 13: This is pretty clever! We did something similar, but your code actually looks a lot cleaner than ours!

EthanAGi (commented):
This looks really cool! I like the idea of using the person class as an in between for any classes that creates a person! That's very clever!

- Define a full name record (first, last, middle) handling optional and null appropriately. ✔
  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Provide a comment on the full name record on why or why not the type is immutable. ✔
- Define an `IEntity` interface: ✔
  - Add an Id property of type `Guid` that is an init-only setter. ✔
  - Add a Name property that is `string`. ✔
- Define an abstract base class that implements `IEntity` - appropriately choosing to implement the interface explicitly or implicitly. ✔
  - Do not implement the `Name` property in this abstract class. ❌ <- I could be incorrect but since you're name variable in the you BaseClass does have a setter, meaning it can be implemented.
  - Do force any derived classes to provide an implementation for `Name`. ✔
- Define book, student, and employee records - all with entity capabilities. ✔
  - Choose a thoughtful approach for the value of `Name` for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ✔
  - Provide a comment on each interface **member** in each entity explaining why you implemented it implicitly or explicitly. ✔
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ✔
  - Given the properties on the full name record, you should consider which entities it makes sense to use it. ✔
  - Test that the equality behavior on these entities behaves as expected. ✔
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
- Unit test the Storage class using various entities to ensure that there are no bugs.
- Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.)
  - Pay special attention to the CreateLogger methods throughout.

## Fundamentals

- Be sure you enable:
  - Enable Nullability for all projects  ✔
  - Ensure that you turn on code analysis for all projects(`EnableNETAnalyzers`) ✔
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
  - Ensure that you turn on Nullability (`Nullable`) ✔
- **Ensure there are no errors or warnings (including code analysis warnings)** ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔

Logger/IEntity.cs
```diff
@@ -2,6 +2,7 @@
 public interface IEntity
 {
 
-        // Place members here.
+    Guid Id { get; init; }
+    string Name { get; set; }
```
EthanAGi, line 6: Sorry, I placed the comment in the wrong area. This is where I'm a bit worried on the setter, as I believe this may allow the name to be implemented in the interface. I could be incorrect on this.

rkdeboer (commented):
Reviewing now

Logger/Book.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Logger;
+
+public record Book : BaseClass
+{
+    public string? Title { get; set; }
+    public string? Author { get; set; }
+    public int YearPublished { get; set; }
+
+    // Calculated property for the full name of the book (Title + Author)
+    public string FullName => $"{Title} by {Author}";
```
rkdeboer, line 11: I like the stringing together of title and author for full name

Logger/Person.cs
```diff
@@ -0,0 +1,14 @@
+﻿
+namespace Logger;
+
+public abstract record Person : BaseClass
+{
+    public string? FirstName { get; set; }
```
rkdeboer, line 6: Could middle name be something to implement?

Logger/Student.cs
```diff
@@ -0,0 +1,12 @@
+﻿
+
+namespace Logger;
```
rkdeboer, line 3: Nice work on the person class I did that too!

rkdeboer:

Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for - - Name (not even one generated by an automatically implemented property) - consider a calculated property. ✔

Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.✔
Unit test the Storage class using various entities to ensure that there are no bugs.✔

Logger.Tests/StorageTests.cs
omar5o9: Thanks for the advice, we ended up adding a lot more test to make sure entities work correctly

Logger/FullName.cs
omar5o9: Thanks :)

Logger/Book.cs
omar5o9: Thanks :)

Logger/IEntity.cs
MannyPoptart: Name here should be fine because we made a new name inside of BaseClass so it would not conflict with this one.

Logger/Person.cs
MannyPoptart: Good catch, we implemented middle name into Person.

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,103 @@
...
+    private readonly Storage storage = new();
+    private readonly Book book = new();
+    private readonly Student student = new();
+    private readonly Employee employee = new();
+    
+    
+
+
+    [Fact]
+    public void Employee_Is_Created_WithProperDepartmentAndID()
```
BenjaminMichaelis, line 16:
```suggestion
    public void Employee_Is_CreatedWithProperDepartmentAndID()
```

Logger/Book.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    // Title, Author, and YearPublished are implemented implicitely because they do not
+    // derive from BaseClass
+    public string? Title { get; set; }
+    public string? Author { get; set; }
+    public int YearPublished { get; set; }
+
+    public string FullName => $"{Title} by {Author}";
+
+    public override string Name { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
BenjaminMichaelis, line 14: Why are you throwing a `NotImplementedException`

Logger/Employee.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Logger;
+
+public record Employee : Person
+{
+
+    // EmployeeId and Department are implemented implicitely because they do not
+    // derive from BaseClass
+    public int EmployeeId { get; set; }
+    public string? Department { get; set; }
+    public override string Name { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
BenjaminMichaelis, line 11: Why are you throwing a `NotImplementedException`

Logger/FullName.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Logger;
+
+public record FullName
+{
+
+    // We decided to handle the full name storage this way to handle middle names being in the name or not
+    // We are allowing the name to be mutable because names can be changed
+    private string fullName;
+
+    public FullName(string fName, string mName, string lName)
```
BenjaminMichaelis, line 10: we don't have to worry about longer names having some performance boost or something. full names are more clear.\
```suggestion
    public FullName(string firstName, string middleName, string lastName)
```

Logger/IEntity.cs
```diff
@@ -2,6 +2,7 @@
 public interface IEntity
 {
 
-        // Place members here.
+    Guid Id { get; init; }
```
BenjaminMichaelis, line 5: Why do we need `init`?

Logger/Student.cs
```diff
@@ -0,0 +1,13 @@
...
+
+public record Student : Person
+{
+    // StudentId is implement implicitely because it does not derive from BaseClass
+    public int StudentId { get; set; }
+    public override string Name { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
BenjaminMichaelis, line 9: Why the `NotImplementedException`

Logger.Tests/StorageTests.cs
BenjaminMichaelis, line 1: There are lots of tests that are missing for each model type

PR #125, opened 2024-02-06
Assignment 4 - Ethan Guerin and Ethan Florko

EthanAGi:
@EthanAGi @EthanFlorko


pja6:
will review

Logger/Name.cs
```diff
@@ -0,0 +1,26 @@
+namespace Logger;
+
+
+public record Name(string FirstName, string? MiddleName, string LastName)
```
pja6, line 4: Wow, this is really cool and well explained - small note I think the Guidelines recommend using record class for clarity
```suggestion
public record class Name(string FirstName, string? MiddleName, string LastName)
```

Logger/Book.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record Book(Guid Id, string Title) : IEntity
+{
+    public string Name => Title; //Using a calculated property for name, this way we can avoid fields.
```
pja6, line 12: Would  a calculated property require some logic in it? This might just be an expression-bodied member, don't think they're the same

Logger/Book.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record Book(Guid Id, string Title) : IEntity
```
pja6, line 10: Curious about your choice to inherit from the entity vs the abstract class - wouldn't that have a similar outcome and force the implementation of name?

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,185 @@
...
+    [InlineData("Ethan", "Alexander", null)]
+    [InlineData("", "Alexander", "Guerin")]
+    [InlineData("Ethan", "Alexander", "")]
+
+    public void Employee_InvalidData_ThrowsError(string firstName, string? middleName, string lastName)
+    {
+        Guid id = Guid.NewGuid();
+        Exception ex = Assert.Throws<ArgumentException>(() => new Employee(id, Name.Create(firstName, middleName, lastName)));
+    }
+
```
pja6, line 184: Thorough!

pja6 (commented):
**Tasks**
Define book, student, and employee records - all with entity capabilities. ✔ // believe this is referring to  inheritance
Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property.  ✔// don't think calculated properties were used - but no backing fields were used either
Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌ // all entities were missing explanation/comments
You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌ // no refactoring/Person class created to handle shared code between student/employee
Test that the equality behavior on these entities behaves as expected. ✔ //believe this was completed in the entity tests - no storage class tests were made

Code review 

**Fundamentals**
Be sure you enable:
All of the above should be unit tested ❌// no storage tests

Overall I think it looks good  - we did some things differently but whether one way is better than the other I couldn't say.  
-Per specifications you are missing some comments and the storage tests.

-I'm also assuming that Name should be implemented in a way that you could refactor into its own class since it's mentioned in the specifications. 

-I don't think you use BaseEntity at all, would assume there is some reason to create it/use for it in the code - wasn't super sure why everything inherited from the interface instead, but maybe you could explain in the code comments

Logger/Book.cs
timmytaters, line 10: Rather than requiring a Guid Id be passed to all of the entities, it may be better to create/assign the Guid in one of the entity classes. Also, I believe that this should derive from BaseEntitiy, not IEntity

Logger/Employee.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record Employee(Guid Id, Name FullName) : IEntity
```
timmytaters, line 9: (Same as Book and employee)
Rather than requiring a Guid Id be passed to all of the entities, it may be better to create/assign the Guid in one of the entity classes. Also, I believe that this should derive from BaseEntitiy, not IEntity

Logger/Name.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public static Name Create(string FirstName, string? MiddleName, string LastName)
```
timmytaters, line 6: Maybe have a more specific name, like CreateName?

timmytaters (commented):
Tasks
Define book, student, and employee records - all with entity capabilities. ❌ No Guid assignment within classes
Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌ // No Comments in entities
You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌ // No refactoring a Person class for shared code like name or GUID
Unit test the Storage class using various entities to ensure that there are no bugs.❌//No storage tests

Fundamentals
Be sure you enable:
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ //latest?
All of the above should be unit tested ❌//No storage or BaseEntity tests

No Extra Credit Attempted

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Logger.Tests;
+
+public class EntityTests
+{
+    [Fact]
+    public void Book_Creation_Success()
```
BenjaminMichaelis, line 13: All classes should be split into files each of their own. BookTests.cs, StudentTests.cs, etc

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,185 @@
...
+    [Theory]
+    [InlineData(null, "Alexander", "Guerin")]
+    [InlineData("Ethan", "Alexander", null)]
+    [InlineData("", "Alexander", "Guerin")]
+    [InlineData("Ethan", "Alexander", "")]
+
+    public void Employee_InvalidData_ThrowsError(string firstName, string? middleName, string lastName)
+    {
+        Guid id = Guid.NewGuid();
+        Exception ex = Assert.Throws<ArgumentException>(() => new Employee(id, Name.Create(firstName, middleName, lastName)));
```
BenjaminMichaelis, line 182: You should additionally test that the param in the ArgumentException is what you expect to make sure the exception is what you expect.

Logger.Tests/NameTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using static Logger.Name;
+using Xunit;
+
+namespace Logger.Tests;
+
+public class NameTests
```
BenjaminMichaelis, line 11:
```suggestion
public class FullNameTests
```

Logger/IEntity.cs
```diff
@@ -1,7 +1,11 @@
...
 
-        // Place members here.
+namespace Logger;
 
-}
+public interface IEntity
+{
+// Place members here
+// Int only setter "Guid"
+Guid Id { get; init; }
```
BenjaminMichaelis, line 9: Why does it need an `init`?

Logger/Name.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(FirstName))
+        {
+            throw new ArgumentException("First name cannot be empty or null.", nameof(FirstName));
```
BenjaminMichaelis, line 10: Nice job with these checks!

PR #129, opened 2024-02-11
Assignment 5 - Cole Stewart and Nathan Marsee

colestewart12:
@colestewart12 @NathanMarsee


Tfelchlin13:
Ill review this one

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            Next = Next
+        };
+        Next = newNode;
+    }
+
+    public void Clear()
+    {
+        //Since the list is singly linked, they will be available for garbage collection
+        Next = this;
```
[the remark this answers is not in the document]
Tfelchlin13: i think they may want more on the comment for if there was  a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected.

Tfelchlin13 (commented):
Overall your code looks good may have to add a larger comment in clear but i like how you did your code it looks nice and clear to read! 

Define the Node class
Extra Credit

Fundamentals

Anna-Mclenko:
Ill review as well

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        //Arrange
+        Node<string> testNode = new("First");
+
+        //Act
+        testNode.Append("Second");
+
+        //Assert
+        Assert.False(testNode.Exists("Third"));
+
```
Anna-Mclenko, line 88: Looks awesome! :) I'd probably test more types, not just int and string

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        //Arrange
+        Node<int> testNode = new(1);
+        testNode.Append(2);
+
+        //Act
+        testNode.Clear();
+
+        //Assert
+        Assert.Equal(testNode.Value, testNode.Next.Value);
```
Anna-Mclenko, line 43: Maybe this will be better?

`Assert.Equal<int>(testNode.Value, testNode.Next.Value);`

GenericsHomework/Node.cs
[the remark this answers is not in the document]
Anna-Mclenko: we did the clear function the same way! The way I would explain it is that it will be collected because the outside reference to the deleted nodes are lost.

Anna-Mclenko (commented):
Instructions

Define the Node class✔

The values should be homogeneous.✔

Extra Credit

Fundamentals

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,33 @@
...
+		<TargetFramework>net8.0</TargetFramework>
+		<IsPackable>false</IsPackable>
+		<Nullable>enable</Nullable>
+		<LangVersion>12.0</LangVersion>
+		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+		<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<EnableNETAnalyzers>true</EnableNETAnalyzers>
+		<IsPackable>false</IsPackable>
```
MattA41, line 13: i would recommend setting the analyzers to the latest to catch more potential errors

MattA41 (commented):
tbh It looks good im not seeing anything that is jumping out to me as bad


Overall your code looks good may have to add a larger comment in clear but i like how you did your code it looks nice and clear to read!

Define the Node class
Extra Credit

Fundamentals

GenericsHomework.Tests/NodeTests.cs
colestewart12, line 43:
```suggestion
        Assert.Equal<int>(testNode.Value, testNode.Next.Value);
```

colestewart12: Thanks!

GenericsHomework/Node.cs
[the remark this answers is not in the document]
colestewart12: Thank you both!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+        Node<T> newNode = new(value)
+        {
+            Next = Next
+        };
+        Next = newNode;
+    }
+
+    public void Clear()
+    {
+        //Since the list is singly linked, they will be available for garbage collection
```
colestewart12, line 38:
```suggestion
        //Since the outside references to the deleted nodes will be lost, they will be available for garbage collection
```

GenericsHomework.Tests/NodeTests.cs
colestewart12: That's a good call thanks

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+        Next = this;
+    }
+
+    public bool Exists(T value)
+    {
+        Node<T> current = this;
+
+        do
+        {
+            bool isFound = current.Value?.Equals(value) ?? throw new InvalidOperationException(nameof(Value));
```
BenjaminMichaelis, line 48: This is checking for the reference, you want an equality comparison.
ex:
` if (EqualityComparer<T>.Default.Equals(node.Value, value)) return true;`

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (isFound) 
+                return true;
```
BenjaminMichaelis, line 50:
```suggestion
            if (Value == value) return true;
```
What if the item supposed to be null? Do we want to throw an exception?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+        Value = value;
+
+        // Points at itself initially
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+        return Value?.ToString() ?? throw new InvalidOperationException($"{nameof(Value)} is null.");
+    }
```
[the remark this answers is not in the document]
BenjaminMichaelis: Dont throw Exception from `ToString` (https://essentialcsharp.com/overriding-object-members#overriding-tostring)

PR #130, opened 2024-02-12
Assignment5 - Anna Mikhailenko and Sergei Ashikhmin 

Anna-Mclenko:
Team: @SergeiAshikhmin and @Anna-Mclenko


Joshua-Lester3:
reviewing :)

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <LangVersion>12.0</LangVersion>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
```
Joshua-Lester3, line 10: As noted in requirements comment, move common properties to props file

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace GenericsHomework.Tests;
+
+public class NodeTests
+{
+
+    [Fact]
+    public void Node_NextPointToItself_Success()
+    {
+        Node<string> node = new("Inigo Montoya");
```
Joshua-Lester3, line 11: I see every test starts with this line. Could be refactored to have 'node' as a field that's set in an initializer/constructor so you don't have to recreate it. But functionally I think the two are the same (because of instance modeling or whatever it's called)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,88 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>
+{
+    private Node<T>? _Next;
```
Joshua-Lester3, line 5: I would argue this shouldn't be nullable because there is no situation where _Next should be allowed to be null -- even if the linked list is of size 1 it points back to itself

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,88 @@
...
+    public T Data { get; set; }
+    public Node(T data)
+    {
+        Data = data;
+        Next = this;
+    }
+    public Node<T> Next {
```
Joshua-Lester3, line 12: If _Next is made to be non-nullable, this would be better implemented as an auto property

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            throw new ArgumentException("The value already exists");
+        }
+
+        Node<T> cur = this;
+
+        while( cur.Next != this )
+        {
+            cur = cur.Next;
+        }
```
Joshua-Lester3, line 36: The requirements say to add the 'data' as a node after the current node. This code looks like it adds the new node before(although if the linked list's size is 1 when you're adding, it functions the same way - any more than a size of 1 and it functions differently). If you delete the while loop I believe it'll follow the requirements

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        Node<T> nextNode = new(data);
+        cur.Next = nextNode;
+        nextNode.Next = this;
+    }
+
+    public override string ToString()
```
Joshua-Lester3, line 43: My interpretation of the requirements list is the ToString method is just for the node, not the linked list. I could be wrong though

Joshua-Lester3 (commented):
Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ✔
Define the Node class ✔
The values should be homogeneous. ✔
Add a ToString() override that writes out the value's ToString() result. ❌ Could be right or wrong depending how they want us to implement it. How I understood this requirement is ToString() just prints the Node's Data field. 
The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌ Next is non-nullable, but the backing field is nullable. I'm not sure if that violates this requirement, but my assumption is it does
Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ❌ The Append method adds the node before the current node instead of after
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌ No VennDiagram structure
Implement System.Collections.Generic.ICollection<T> on the Node class ❌ Did not implement ICollection<T>
Fundamentals
Place all shared project properties into a Directory.Build.Props file. ❌ No Directory.Build.props file created with shared properties
Place all shared project items into a Directory.Build.targets file. ✔
For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌ I don't think AreEqual is in Xunit, but generics weren't used

Tvor369:
Ill add another review...

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        Assert.False(node.Exists("Princes Bumble Gum"));
+    }
+
+    [Fact]
+    public void ToString_PrintList_Success()
+    {
+        Node<string> node = new("Inigo Montoya");
+        node.Append("Butter Cup");
+        node.Append("Prince Johan");
+        Assert.Equal("Linked List: Node 1: Inigo Montoya, Node 2: Butter Cup, Node 3: Prince Johan, }", node.ToString());
```
Tvor369, line 64: I like how clear you guys made the ToString, how each value is matched with a node number

GenericsHomework/Node.cs
Tvor369, line 5: I would 2nd Joshua's comment here, the default node that Next points to is itself, like how you have in the body of Node.

GenericsHomework/Node.cs
Tvor369, line 43: We also interpreted it as returning the ToString of a value of an individual node, but this is cooler

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,88 @@
...
+            cur = cur.Next;
+        }while(cur != this);
+
+        outPut += "}";
+        return outPut;
+    }
+
+    public bool Exists(T data)
+    {
+        bool result = false;
```
Tvor369, line 61: You could get rid of the bool variable _result_ and just have it either return true in that if statement or return false otherwise at the end, since we only have to find the first instance of a duplicate. But I also understand it could be more readable this way.

Tvor369:
### Instructions
- Create a class library project called "GenericsHomework.". ✔
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.✔
- Define the Node class✔
- The values should be homogeneous.✔
- Include a constructor that takes a value. (No validation is necessary on the value). ✔
- Add a ToString() override that writes out the value's ToString() result. ❌✔ (Idk if they wanted it your way or just the single node)
- Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
- The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌
- The Next property setter should be private. ✔
- Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ❌(Joshua's interpretation might be right on this, we did it wrong as well if thats the case.)

### Fundamentals
- Place all shared project properties into a Directory.Build.Props file.✔
- nullable reference types is enabled ✔
- Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
- Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
- For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌ (We didn't either)

GenericsHomework/Node.cs
Anna-Mclenko: oh right thanks for catching that! we'll fix that

Anna-Mclenko: oh right thanks for catching that! we'll fix that

GenericsHomework/Node.cs
Anna-Mclenko: ooooooh right!!!! fixed that!

GenericsHomework/Node.cs
Anna-Mclenko: oh... I guess we missundersdtood. we that since the word "append" was used that the node needs to be at the "end" of the list. thanks for catching that. we fixed that

GenericsHomework/Node.cs
Anna-Mclenko: since we are working with a linked list it doesn't make sense to just print the data in only one node. Thus, when the to string method is called on the first node we decided to print the entire list

Anna-Mclenko: since we are working with a linked list it doesn't make sense to just print the data in only one node. Thus, when the to string method is called on the first node we decided to print the entire list

GenericsHomework/Node.cs
Anna-Mclenko: thats a valid point! thanks for catching that!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public void Clear()
+    {
+        //Yur do not need to worry about Garbage collection if implemented like below.
+        //The reason for this is because there does not exist a root reference to the existing object anymore.
+        //Thus, the garbage collector will automatically collect the other nodes.
+        Node<T> curr = this;
+
+        curr.Next = this;
```
BenjaminMichaelis, line 72:
```suggestion
        Next = this;
```
can't you just do this?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public bool Exists(T data)
+    {
+
+        Node<T> cur = this;
+
+        do
+        {
+            if (cur.Data!.Equals(data))
```
BenjaminMichaelis, line 53: How do you know Data is not going to be null? What if cur.Data and data are both null? are they equal then and should return true?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        do{
+            count++;
+            outPut += $"Node {count}: {cur.Data}, ";
+            cur = cur.Next;
+        }while(cur != this);
+
+        outPut += "}";
+        return outPut;
+    }
```
BenjaminMichaelis, line 44:
```suggestion
    public override string? ToString()
    {
        if (Data== null) return null;
        return Data.ToString();
    }
```
I think a simpler approach. When I call `ToString` on a node, I'm not sure I expect to get the entire list in string form

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        Assert.NotEqual(node.Data, node.Next.Data);
+    }
+
+    [Fact]
+    public void Node_NextPointToTheLastNode_Success()
+    {
+        Node<double> node = new(1.0);
+        node.Append(2.0);
+        node.Append(3.0);
+        Assert.Equal(2.0, node.Next.Next.Data);
```
BenjaminMichaelis, line 29:
```suggestion
        Assert.Equal(3.0, node.Next.Next.Next.Data);
```

PR #131, opened 2024-02-12
Assignment5- Omar Facundo and Manny Aguilera

omar5o9:
@omar5o9 and @MannyPoptart


colestewart12 (commented):
**Instructions**
Define the Node class✔
The values should be homogeneous.✔
The Next property setter should be private. ❌
**Fundamentals**
nullable reference types is enabled ❌
Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ❌
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌
and enabled .NET analyzers for both projects ❌

Build is successful and tests are running successfully. A couple of suggestions including adding properties to csproj files but overall looks pretty good!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+namespace GenericsHomework;
+
+
+public class Node<T>
+{
+
+    public Node<T> Next { get; set; }
+    public T Data { get; set; }
```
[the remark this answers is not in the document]
robertogarcia11: +1 on this! theres really no need for a setter in this case.

GenericsHomework.Tests/NodeTests.cs
robertogarcia11, line 1: other than nitpicky test method names, tests look good!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+        Node<T> cur = this;
+        while (cur.Next != this)
+        {
+            if (cur.Data != null && cur.Data.Equals(value))
+            {
+                return true;
+            }
+            cur = cur.Next;
+        }
+        return cur.Data != null && cur.Data.Equals(value);
```
robertogarcia11, line 69: my partner and I used a do-while loop instead of a regular while loop and your code looks a lot more efficient and simple! nice job!

robertogarcia11: might steal idea, tehe

robertogarcia11:
Only minor changes are needed but overall great work! both of you!

Instructions
Define the Node class✔
The values should be homogeneous.✔
The Next property setter should be private. ❌
Fundamentals

Directory.Build.Props
```diff
@@ -0,0 +1,11 @@
+﻿<Project>
+
+  <PropertyGroup>
+	<LangVersion>12.0</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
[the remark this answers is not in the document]
omar5o9: Thanks added 👍

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <IsPackable>false</IsPackable>
```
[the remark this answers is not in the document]
omar5o9: Got it boss

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+namespace GenericsHomework.Tests;
+
+public class NodeTests
+{
+
+    [Fact]
+    public void DuplicateValueTest()
```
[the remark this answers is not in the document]
omar5o9: Okay thanks, changed all the test names

GenericsHomework/Node.cs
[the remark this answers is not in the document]
MannyPoptart: Noted, good catch.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+            result += cur.Data + " -> ";
+            cur = cur.Next;
+        }
+        result += cur.Data;
+        return result;
+    }
+    
+    public void Clear()
+    {
+        // C# will automatically garbage collect the rest of the nodes
```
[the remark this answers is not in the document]
MannyPoptart: Thanks.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+        Data = data;
+        Next = this;
+    }
+
+    public void Append(T value)
+    {
+        if(Exists(value))
+        {
+            throw new InvalidOperationException(nameof(value));
+        }
```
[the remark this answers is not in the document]
MannyPoptart: Thanks.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+namespace GenericsHomework;
+
+
+public class Node<T>
+{
+
+    public Node<T> Next { get; set; }
```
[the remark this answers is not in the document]
MannyPoptart: Thanks for the catch, we have set it to private now. 👍

Directory.Build.targets
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <Target Name="Target" AfterTargets="Build">
+      <Message Text="yo" Importance="high" />
```
BenjaminMichaelis, line 3: Lol

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,33 @@
...
+	<Nullable>enable</Nullable>
+	<LangVersion>12.0</LangVersion>
+	<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+	<ImplicitUsings>enable</ImplicitUsings>
+	<EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<IsTestProject>true</IsTestProject>
+  </PropertyGroup>
```
BenjaminMichaelis, line 14:
```suggestion
    <IsPackable>false</IsPackable>
	<IsTestProject>true</IsTestProject>
  </PropertyGroup>
```
Since these are in your .props file, you shouldn't need them here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+    public void Append(T value)
+    {
+        if(Exists(value))
+        {
+            throw new InvalidOperationException(nameof(value));
+        }   
+        Node<T> newNode = new(value);
+        Node<T> cur = this;
+
+        while (cur.Next != this)
```
BenjaminMichaelis, line 26: What if you only have one node in the list? Couldn't this be an infinite loop then?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            cur = cur.Next;
+        }
+
+        newNode.Next = cur.Next;
+        cur.Next = newNode;
```
BenjaminMichaelis, line 32:
```suggestion
  if (Exists(value)) throw new ArgumentException("This item already exists in the list");
  Node<T> newNode = new(value);
  newNode.Next = Next;
  Next = newNode;
  return newNode;
```
I think this can be simplified

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        Node<T> cur = this;
+        string result = "";
+        while (cur.Next != this)
+        {
+            result += cur.Data + " -> ";
+            cur = cur.Next;
+        }
+        result += cur.Data;
+        return result;
```
BenjaminMichaelis, line 45:
```suggestion
        if (Value == null) return null;
        return Value.ToString();
```
I think this can be simplified. When I to string a node, I don't think I want to get the whole "list"

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+    }
+    
+    public void Clear()
+    {
+        // C# will automatically garbage collect the rest of the nodes
+        Node<T> cur = this;
+        cur.Next = this;
```
BenjaminMichaelis, line 52:
```suggestion
        Next = this;
```
even simpler

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    }
+
+    public bool Exists(T value)
+    {
+
+        Node<T> cur = this;
+        while (cur.Next != this)
+        {
+            if (cur.Data != null && cur.Data.Equals(value))
```
BenjaminMichaelis, line 63: What if Data is null and value is null? Aren't they equal then and then should return true?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return true;
+            }
+            cur = cur.Next;
+        }
+        return cur.Data != null && cur.Data.Equals(value);
```
BenjaminMichaelis, line 69:
```suggestion
        do
        {
            if (EqualityComparer<T>.Default.Equals(node.Value, value)) return true;
            node = node.Next;
        } while (node != this);
        return false;
```
A do while loop would reduce redundant code and logic

PR #132, opened 2024-02-12
Assignment5 - Ethan Guerin and Ethan Florko

EthanAGi:
@EthanAGi @EthanFlorko


JohanneJayde:
Reviewing!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    //Can add contraint such as "where T : class" to ensure homogeneous values
+    //Can possibly also add "public T Data { get; }" to ensure pproperty is non nullable.
+    public T Data { get; } = data;
+    private Node<T>? next; 
+    public Node<T> Next
+    {
+        get => next ?? this;
+        private set => next = value;
+    }
```
JohanneJayde, line 13:
```suggestion
public class Node<T>
{
    //Can add contraint such as "where T : class" to ensure homogeneous values
    //Can possibly also add "public T Data { get; }" to ensure pproperty is non nullable.
    public T Data { get; };
    public Node<T> Next {get; private set;}
    public Node(T data){
            Next = this;
            Data = data;
    }
```
Having a backing field is unnecessary and overcomplicated things. Just use a regular constructor instead of primary which will allow you to set Next to "this" instead of having a private backing field.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>(T data)
+{
+    //Can add contraint such as "where T : class" to ensure homogeneous values
```
JohanneJayde, line 5: Having "where T : class" doesn't make the data homogeneous. That will just limit what types you can utilize in the Node class.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+        else
+        {
+            Node<T> cur = Next;
+            while(cur.Next != this)
+            {
+                cur = cur.Next;
+            }
+            cur.Next = newNode;
+        }
```
JohanneJayde, line 39: This is not being tested currently. Right now you have one test for appends which will only cover the if-statement where there is only one element. Consider adding a test that will check when multiple elements are being appended to the Node

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
...
+                return true;
+            }
+            cur = cur.Next;
+        } while (cur != this);
+        return false;
+    }
+
+    public void Clear()
+    {
+        Node<T> cur = this;//Create a new Node that holds the same data, this node is not connected to the rest of the list.
```
JohanneJayde, line 59:
```suggestion
        //Create a new Node that holds the same data, this node is not connected to the rest of the list.
```
No need for cur in this case

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    [Fact]
+    public void Append_DuplicateNull_ThrowsError()
+    {
+        Node<string> node = new(null!);
+        node.Append("second");
+
+        Exception exception = Assert.Throws<ArgumentException>(() => node.Append(null!));
+        Assert.Contains("Duplicate Value cannot be added.", exception.Message);
+    }
+}
```
JohanneJayde, line 116: These tests are just repeats of previous tests. They result in the same behavior/result as the non-null tests. I would suggest making each of these tests data-driven remove the null tests. I have provided an example in my suggestions

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        Assert.True(object.ReferenceEquals(node, node.Next));
+    }
+
+    [Fact]
+    public void Exists_ValueExists_ReturnsTrue()
+    {
+        Node<int> node = new(1);
+        bool exists = node.Exists(1);
+        Assert.True(exists);
+    }
```
JohanneJayde, line 51:
```suggestion
    [Theory]
    [InlineData("Hi")]
    [InlineData(null!)]
    public void Exists_ValueExists_ReturnsTrue(string data)
    {
        Node<string> node = new(data);
        bool exists = node.Exists(data);
        Assert.True(exists);
    }
```
Here's an example for making the Null and non-null test one test. Exists for non-null and null data is doing the same thing so only one test is necessary in my opinion

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    [Fact]
+    public void Append_Null_ValueNull()
+    {
+        Node<string> node = new("first");
+        node.Append(null!);
+
+        Assert.Null(node.Next.Data);
+    }
```
JohanneJayde, line 84: This test isn't doing much. All it is doing is appending null and then checking if the value is null. This is part of a much larger issue. You have made Data non-nullable within the Node class. But this test is saying by passing null into the constructor Data should be null. So they are constricting each other. It makes more sense for this test to throw an ArgumentNullException since the Data is non-nullable. It seems like you want Data to be nullable based on the tests that have been written but maybe not.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    [Fact]
+    public void Exists_NullNotThere_ReturnsFalse()
+    {
+        Node<string> node = new("first");
+        node.Append("second");
+
+        bool exists = node.Exists(null!);
+
+        Assert.False(exists);
```
JohanneJayde, line 104:
```suggestion
        Assert.False(node.Exists(null!));
```
There is no need to define exists variable

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    [Fact]
+    public void Exists_NullValue_ReturnsTrue()
+    {
+        Node<string> node = new(null!);
+        node.Append("second");
+
+        bool exists = node.Exists(null!);
+        Assert.True(exists);
```
JohanneJayde, line 93:
```suggestion
        Assert.True(node.Exists(null!));
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
...
+        newNode.Next = this;
+    }
+
+    public bool Exists(T data)
+    {
+        Node<T> cur = this;
+        do
+        {
+            if(EqualityComparer<T>.Default.Equals(cur.Data, data))
+            {
```
JohanneJayde, line 49: Cool use of EqualityComparrer<T>.Default. I haven't seen this before but it makes sense why it can be better than Equals() when using generics.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    [Fact]
+    public void ToString_ReturnsCorrect()
+    {
+        Node<int> node = new(1999);
+        string res = node.ToString();
+        Assert.Equal("1999", res);
+    }
+
```
JohanneJayde, line 21: Your toString method returns an empty string if null is passed. There is no test checking for that

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        Node<int> node = new(1);
+        Assert.True(object.ReferenceEquals(node, node.Next));
+    }
+
+    [Fact]
+    public void Exists_ValueExists_ReturnsTrue()
+    {
+        Node<int> node = new(1);
+        bool exists = node.Exists(1);
+        Assert.True(exists);
```
JohanneJayde, line 50:
```suggestion
        Assert.True(node.Exists(1));
```

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+	  <OutputType>Library</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+	  <LangVersion>latest</LangVersion>
```
JohanneJayde, line 8: That's interesting. I didn't know could specify it as latest for `LangVersion`

JohanneJayde (commented):
# Instructions
Define the Node class
# Extra Credit

# Fundamentals
Place all shared project properties into a Directory.Build.Props file. ❌
Place all shared project items into a Directory.Build.targets file. - `Doesn't apply to current Assignment`
All of the above should be unit tested ❌ - `Append method is missing test for when you are appending elements after the second one`

MattA41:
reviewing

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>(T data)
+{
+    //Can add contraint such as "where T : class" to ensure homogeneous values
+    //Can possibly also add "public T Data { get; }" to ensure pproperty is non nullable.
+    public T Data { get; } = data;
+    private Node<T>? next;
```
MattA41, line 8: part of the assignment states that node next needs to be non-nullable but right here you are making it null

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    [Fact]
+    public void Append_DuplicateValue_ThrowsError()
+    {
+        Node<int> node = new(1);
+        node.Append(2);
+
+        Exception exception = Assert.Throws<ArgumentException>(() => node.Append(2));
+        Assert.Contains("Duplicate Value cannot be added.", exception.Message);
+    }
```
MattA41, line 68: is there a test for when appending a list with more than one node

MattA41 (commented):
        Define the Node class
        Add a ToString() override that writes out the value's ToString() result.✔
            The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌

Extra Credit

    Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals

    nullable reference types is enabled ❌ - I don't see this file

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    [Fact]
+    public void Append_DuplicateValue_ThrowsError()
+    {
+        Node<int> node = new(1);
+        node.Append(2);
+
+        Exception exception = Assert.Throws<ArgumentException>(() => node.Append(2));
+        Assert.Contains("Duplicate Value cannot be added.", exception.Message);
```
BenjaminMichaelis, line 67: Nice job checking something on the exception itself!

GenericsHomework.Tests/NodeTests.cs
BenjaminMichaelis: Great point @JohanneJayde.

I think Data should not be nullable by default. But I could still declare it as nullable when I instantiate it, by doing like `Node<string?> node = new("thing")` which I think is OK, and then `Append(null);`

However, you are totally right that this is a strange test, the purpose is not clear, because you are forcing null in. This test doesn't add value other than you are showing that you can "turn off" null warnings by doing the null-forgiving operator

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        Node<string> node = new("first");
+        node.Append("second");
+
+        bool exists = node.Exists(null!);
+
+        Assert.False(exists);
+    }
+
+    [Fact]
+    public void Append_DuplicateNull_ThrowsError()
```
BenjaminMichaelis, line 107: Why is `null!` being used in so many places? This just seems like a bad practice. It should only be used if you are testing what happens IF you force null to go in. using just a string in many of these cases is sufficient and I would argue better.

GenericsHomework/GenericsHomework.csproj
BenjaminMichaelis: https://learn.microsoft.com/dotnet/csharp/language-reference/compiler-options/language?WT.mc_id=8B97120A00B57354#langversion

It is the same as 12.0 rn, but will mean you automatically grab 13.0 when that releases. It's an intentional choice when to use either. In most cases I want to intentionally upgrade, but there is a place for both

GenericsHomework/Node.cs
BenjaminMichaelis: I agree, what if I want to do `node<int>`?

GenericsHomework/Node.cs
BenjaminMichaelis: next should point to itself until it has something else to point to

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+    //Can add contraint such as "where T : class" to ensure homogeneous values
+    //Can possibly also add "public T Data { get; }" to ensure pproperty is non nullable.
+
+    public T Data { get; } = data;
+    private Node<T>? next; 
+    public Node<T> Next
+    {
+        get => next ?? this;
+        private set => next = value;
+    }
```
BenjaminMichaelis, line 14:
```suggestion
    public Node<T> Next { get; private set; }
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public override string ToString()
+    {
+        return Data?.ToString() ?? string.Empty;
+    }
```
BenjaminMichaelis, line 19:
```suggestion
    public override string? ToString()
    {
        if (Value == null) return null;
        return Value.ToString();
    }
```

BenjaminMichaelis: DO NOT return an empty string or null from ToString(). (https://essentialcsharp.com/overriding-object-members#overriding-tostring)

PR #133, opened 2024-02-13
Assignment5 - Paulson Ray Thompson & Trevor Stewart

Purplepigs0129:
Trevor's Git - @Tvor369

### DIFF
--- a/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework.Tests/NodeTests.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GenericsHomework.Tests;
+
+//Reminder: For this assignment, always use Assert.AreEqual<T>() (the generic version)
+
+public class NodeTests
+{
+    
+}
+
--- a/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework.Tests/NodeTests.cs
@@ -8,14 +8,14 @@ public class NodeTests
 {
     [Fact]
     public void Node_OneNodeLoopsOnItself_Equal()
-    {
+    {//Node not recognized
         Node<int> node = new(3);
         Assert.Equal(node.Data, node.Next.Data);
     }
 
     [Fact]
     public void Node_MultipleNodes_NotEqual()
-    {
+    {   
         Node<int> node = new(3);
         node.Append(5);
         Assert.NotEqual(node, node.Next);
--- a/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework.Tests/NodeTests.cs
@@ -1,8 +1,3 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace GenericsHomework.Tests;
--- a/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework.Tests/NodeTests.cs
@@ -2,19 +2,25 @@ using Xunit;
 
 namespace GenericsHomework.Tests;
 
-//Reminder: For this assignment, always use Assert.AreEqual<T>() (the generic version)
-
 public class NodeTests
 {
     [Fact]
-    public void Next_OneNodeLoopsOnItself_Equal()
-    {//Node not recognized
+    public void Next_NodeLoopsOnItself_Equal()
+    {
         Node<int> node = new(3);
         Assert.Equal(node.Data, node.Next.Data);
     }
 
     [Fact]
-    public void Append_AddNodeWithDiffValue_NotEqual()
+    public void Append_Node_Equal()
+    {
+        Node<int> node = new(6);
+        node.Append(7);
+        Assert.Equal(7, node.Next.Data);
+    }
+
+    [Fact]
+    public void Append_NodeWithDifValue_NotEqual()
     {
         Node<int> node = new(3);
         node.Append(5);
@@ -30,7 +36,7 @@ public class NodeTests
     }
 
     [Fact]
-    public void Exists_CheckValueThatIsInList_True()
+    public void Exists_ValueInList_ReturnsTrue()
     {
         Node<string> node = new("apple");
         node.Append("pear");
@@ -39,7 +45,7 @@ public class NodeTests
     }
 
     [Fact]
-    public void Exists_CheckValueThatIsNotInList_False()
+    public void Exists_ValueInList_ReturnsFalse()
     {
         Node<string> node = new("apple");
         node.Append("pear");
@@ -48,7 +54,7 @@ public class NodeTests
     }
 
     [Fact]
-    public void Append_TryToAppendDuplicate_ThrowException()
+    public void Append_AppendsDuplicate_ThrowsException()
     {
         Node<int> node = new(3);
         node.Append(5);
@@ -57,7 +63,7 @@ public class NodeTests
     }
 
     [Fact]
-    public void Append_TryToAppendNullValue_ThrowException()
+    public void Append_NullValue_ThrowsException()
     {
         Node<string> node = new("apple");
         node.Append("pear");
@@ -66,7 +72,7 @@ public class NodeTests
     }
 
     [Fact]
-    public void Clear_CreateListAndClear_HeadShouldRemain()
+    public void Clear_ClearsNewList_HeadRemains()
     {
         Node<string> node = new("apple");
         node.Append("pear");
@@ -90,8 +96,6 @@ public class NodeTests
         Assert.True(node.Exists("apple"));
         Assert.False(node.Exists("pear"));
         Assert.False(node.Exists("orange"));
-
-        //TODO: add tests for ToString
     }
 
     [Fact]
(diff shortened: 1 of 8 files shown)


alyxadarr:
reviewing :)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        Node<T> cur = this; //new node holds the same data while disconnected from the rest of the list
+        cur.Next = this; //next will remove the rest of the references
+    }
+
+    public bool Exists(T value)
+    {
+        //Search list to see if value exists
+        Node<T> cur = this;
+        do
```
alyxadarr, line 51: good job we did the same thing here!

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    public void Exists_CheckValueThatIsNotInList_False()
+    {
+        Node<string> node = new("apple");
+        node.Append("pear");
+        node.Append("orange");
+        Assert.False(node.Exists("cheese"));
+    }
+
+    [Fact]
+    public void Append_TryToAppendDuplicate_ThrowException()
```
alyxadarr, line 51: i feel like you could make your test method names a little more shorter and easier for readability.

alyxadarr, line 51:
```
Append_DuplicateValue_ThrowsException
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        Node<int> node = new(3);
+        node.Append(5);
+        node.Append(7);
+        Assert.Throws<ArgumentException>(() => node.Append(3));
+    }
+
+    [Fact]
+    public void Append_TryToAppendNullValue_ThrowException()
```
alyxadarr, line 60:
```
Append_NullValue_ThrowsException
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    [Fact]
+    public void Next_LoopsAroundToSelf_Equal()
+    {
+        Node<int> node = new(3);
+        node.Append(5);
+        Assert.Equal(node, node.Next.Next);
+    }
+
+    [Fact]
+    public void Exists_CheckValueThatIsInList_True()
```
alyxadarr, line 33:
```
Exists_ValueInList_ReturnsTrue
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+public class NodeTests
+{
+    [Fact]
+    public void Next_OneNodeLoopsOnItself_Equal()
+    {//Node not recognized
+        Node<int> node = new(3);
+        Assert.Equal(node.Data, node.Next.Data);
+    }
+
+    [Fact]
```
alyxadarr, line 16: i think you could use a Test that successfully checks that the Append method works and adds nodes. I know you have a Append_AddNodeWithDiffValue_NotEqual() method but you could do something like this 
Node<string> newNode = new("Hi");
newNode.Append("You rock");
Assert.Equal("You rock", newNode.Next.Data);
this could really test for the append method functionality

alyxadarr (commented):
Great job guys! Overall your code looks clean, efficient and simple over complex. Create a class library project called "GenericsHomework.". ✔
Define the Node class

EthanAGi:
I'll add another review

EthanAGi (commented):
This is quite good! The first review pickled up on a lot, I feel I don't have too much to add. Good job you guys! This is really well done!

  - Define the `Node` class

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals

## See [Docs](https://github.com/IntelliTect-Samples/EWU-CSCD371-2024-Winter/tree/main/Docs)

GenericsHomework.Tests/NodeTests.cs
Purplepigs0129: Definitely will update the names on the methods so it doesn't take as much time to read and understand what the test is testing.

Purplepigs0129: Thanks!

GenericsHomework.Tests/NodeTests.cs
Purplepigs0129: Thanks

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    public void Exists_CheckValueThatIsInList_True()
+    {
+        Node<string> node = new("apple");
+        node.Append("pear");
+        node.Append("orange");
+        Assert.True(node.Exists("pear"));
+    }
+
+    [Fact]
+    public void Exists_CheckValueThatIsNotInList_False()
```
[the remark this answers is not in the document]
Purplepigs0129: Thanks!

GenericsHomework.Tests/NodeTests.cs
Purplepigs0129: Thanks!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        ArgumentNullException.ThrowIfNull(Data);
+        return Data.ToString()!;
+    }
+
+    public void Append(T value)
+    {
+        if (value == null) throw new ArgumentNullException(nameof(value));//Prevent null values from being added
+
+        //Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case)
+        if (Exists(value) == true) throw new ArgumentException("Cannot have duplicate: " + nameof(value));
```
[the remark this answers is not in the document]
Purplepigs0129: Thanks it was originally just a exception but got an error on it due to it being vague so this is what happened next

GenericsHomework/Node.cs
Purplepigs0129: Thanks!

GenericsHomework.Tests/NodeTests.cs
Purplepigs0129: We are at 100% code coverage already since this is technically being tested by our other methods I don't see the need to add it. However, another test won't hurt and it certainly makes sense to test this along with the others when creating the tests to test the Append method.

BenjaminMichaelis: 100% code coverage doesn't actually mean your code is fully tested. Every line is passed over during a test, but doesn't mean every case is tested. Mutation testing can help expose this. https://stryker-mutator.io/

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,115 @@
...
+    [Fact]
+    public void Append_Node_Equal()
+    {
+        Node<int> node = new(6);
+        node.Append(7);
+        Assert.Equal(7, node.Next.Data);
+    }
+
+    [Fact]
+    public void Append_NodeWithDifValue_NotEqual()
```
BenjaminMichaelis, line 23:
```suggestion
    public void Append_NodeWithDifferentValue_NotEqual()
```
No need to use shorthand

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        Next = this;//else refers back to itself if there are no other nodes in the list.
+    }
+
+
+    public override string ToString()
+    {
+        //Add a ToString() override that writes out the value's ToString() result.
+        ArgumentNullException.ThrowIfNull(Data);
+        return Data.ToString()!;
+    }
```
BenjaminMichaelis, line 20:
```suggestion
    public override string? ToString()
    {
        if (Value == null) return null;
        return Value.ToString();
    }
```

BenjaminMichaelis: Dont throw Exception from `ToString` (https://essentialcsharp.com/overriding-object-members#overriding-tostring)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    public void Append(T value)
+    {
+        if (value == null) throw new ArgumentNullException(nameof(value));//Prevent null values from being added
```
BenjaminMichaelis, line 24: What if `T` is nullable?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        Node<T> newNode = new(value);
+        cur.Next = newNode;
+        newNode.Next = this;//Loop the newNode back to the head
+    }
+
+    public void Clear() //c# will auto collect garbage for the rest of the nodes
+    {
+        Node<T> cur = this; //new node holds the same data while disconnected from the rest of the list
+        cur.Next = this; //next will remove the rest of the references
```
BenjaminMichaelis, line 44:
```suggestion
        Next = this;
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public bool Exists(T value)
+    {
+        //Search list to see if value exists
+        Node<T> cur = this;
+        do
+        {
+            if (cur.Data!.Equals(value)) return true; //Using ! since Append checks if value is null
```
BenjaminMichaelis, line 53: What if Data and value are both null? Aren't they equal then?

PR #134, opened 2024-02-13
Assignment5 - Josh Pedersen & Conner Hanson

Joshua-Pedersen:
Partner: @connerhanson


SergeiAshikhmin:
I will review this

SergeiAshikhmin (commented):
Code looks good and clean. Good job 👍

Instructions
Define the Node class

I did not see any mentions of extra credit so I assume you skipped it.

Fundamentals

Matthew-Powers:
I will review this

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        // To add once Exists method is here
+        if (Exists(value)) 
+        {
+            throw new System.ArgumentException("Duplicate Value");
+        }
+         
+
+        Node<T> newNode = new(value);
+        Node<T> cur = this;
```
Matthew-Powers, line 30: cur isn't a very descriptive variable name. I'm assuming it's short for 'current'. If so, replace cur with current

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    public override string ToString()
+    {
+            string contents = $"{Value}";
+            return contents;
+    }
+
+    public void Append(T value)
+    {
+        // To add once Exists method is here
```
Matthew-Powers, line 22: Exists method is here, don't need this comment any more

GenericsHomework/Node.cs
Matthew-Powers: Even better might be currentNode

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+        if (Exists(value)) 
+        {
+            throw new System.ArgumentException("Duplicate Value");
+        }
+         
+
+        Node<T> newNode = new(value);
+        Node<T> cur = this;
+
+        while (cur.Next != this) { cur = cur.Next; }
```
Matthew-Powers, line 32: This probably should be broken up into multiple lines for readability
```suggestion
        while (cur.Next != this) 
        { 
            cur = cur.Next; 
        }

```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+        // base case if list is a single element
+        if(this.Next == this)
+        {
+            return;
+        }
+
+        // Reference of next node before clearing 
+        Node<T> curNode = this;
+
+        // Loop through list until your at the node before this
```
Matthew-Powers, line 49: nitpick: you're*

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    public Node(T value)
+    {
+        Value = value;
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+            string contents = $"{Value}";
```
Matthew-Powers, line 16: Kind of weird spacing going on here. Should be reduced to 1 tab

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>
```
Matthew-Powers, line 3:
```suggestion
public class Node<T> : System.Collections.Generic.ICollection<T>

```

Directory.Build.Props
```diff
@@ -0,0 +1,10 @@
+﻿<Project>
+	<PropertyGroup>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<LangVersion>12.0</LangVersion>
+		<Nullable>enable</Nullable>
```
Matthew-Powers, line 5: <TargetFramework>net8.0</TargetFramework>

Matthew-Powers, line 5: Missing \<TargetFramework>net8.0\</TargetFramework>

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>net8.0</TargetFramework>
```
Matthew-Powers, line 4: Can remove this after putting it in the props file

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        // Arrange
+        Node<int> node = new(1);
+        node.Append(2);
+        node.Append(3);
+
+
+        // Act and Assert
+        Assert.Throws<System.ArgumentException>(() => node.Append(1));
+        Assert.Throws<System.ArgumentException>(() => node.Append(3));
```
Matthew-Powers, line 95: Should probably only use a single assert

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>net8.0</TargetFramework>
```
Matthew-Powers, line 4: Can remove this after adding it to props file

GenericsHomework/Node.cs
Matthew-Powers: For extra credit

Matthew-Powers (commented):
Instructions

    Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ❌
--Nodes don't point back to the first item (might not actually be an issue, since this is kind of vague in terms of meaning)
        Define the Node class ✔
        The values should be homogeneous. ✔

Extra Credit

    Each circle contains n items and each item can belong to one or more Circle instances. ❌
    You are not required to use a Node from earlier in the homework for your Venn diagram implementation. ❌
    You are welcome to use existing BCL generic classes for the extra credit. ❌

Fundamentals

    Place all shared project properties into a Directory.Build.Props file. ❌
--Missing: <TargetFramework>net8.0</TargetFramework>in props file
    Place all shared project items into a Directory.Build.targets file. ❌
--File does not exist
--Never used in test, and I see a lot of Assert.Equal

Overall looks pretty good. Just needs some tweaking.

Would also recommend looking into cleaning up your whitespace. EG: Line 18 in NodeTests

GenericsHomework/Node.cs
connerhanson: Changed all instances of cur or curNode to currentNode.  Naming it cur is just a habit I got from doing linked lists in for data structures, thanks for the advice!

GenericsHomework/Node.cs
connerhanson: Fixed, thanks for the reminder!

GenericsHomework/Node.cs
connerhanson: Fixed, Good suggestion!

GenericsHomework.Tests/NodeTests.cs
connerhanson: Here I used two asserts to make sure the exist method loops properly.  First making sure it finds the first node, and second making sure it finds the last node.

connerhanson:
@Matthew-Powers  For this assignment, always use Assert.AreEqual() (the generic version) ❌ --Never used in test, and I see a lot of Assert.Equal 

In xunit Assert.Equal is generic by default.  Found in this document: https://xunit.net/docs/comparisons

GenericsHomework/Node.cs
Joshua-Pedersen: Good catch lol

GenericsHomework/Node.cs
Joshua-Pedersen: Implementation of ICollection will make the code messy, we're choosing to stick to the base tasks.

Directory.Build.Props
Joshua-Pedersen: We moved the ``<TargetFramework>`` into each of the csproj files instead. Placing the ``<TargetFramework>`` in the ``.props`` file causes a fatal error where neither the BuildTest nor certain operating systems are able to build the code.

GenericsHomework/GenericsHomework.csproj
Joshua-Pedersen: We moved the <TargetFramework> into each of the csproj files instead. Placing the <TargetFramework> in the .props file causes a fatal error where neither the BuildTest nor certain operating systems are able to build the code.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,189 @@
...
+    public void ToString_HasValue_ReturnsString()
+    {
+        // Arrange
+        int val =  42;
+
+        // Act
+        Node<int> testNode = new(val);
+
+        // Assert
+        Assert.Equal($"{val}", testNode.ToString());
```
BenjaminMichaelis, line 31:
```suggestion
        Assert.Equal(val.ToString(), testNode.ToString());

```
Why not `.ToString()`?

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,10 @@
...
+	</PropertyGroup>
+	<ItemGroup>
+		<None Remove="Directory.Build.Props"/>
+	</ItemGroup>
```
BenjaminMichaelis, line 8: What's this for?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        Value = value;
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+        string contents = $"{Value}";
+        return contents;
+    }
```
BenjaminMichaelis, line 18:
```suggestion
    public override string? ToString()
    {
        if (Value == null) return null;
        return Value.ToString();
    }

```

slightly improved

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+        Node<T> newNode = new(value);
+        Node<T> currentNode = this;
+
+        while (currentNode.Next != this) 
+        { 
+            currentNode = currentNode.Next; 
+        }
+
+        currentNode.Next = newNode;
+        newNode.Next = this;
```
BenjaminMichaelis, line 37:
```suggestion
         
        Node<T> newNode = new(value)
        {
            Next = Next
        };
        Next = newNode;
        return newNode;
```
bit simpler logic

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+        while (currentNode.Next != this)
+        {
+            currentNode = currentNode.Next;
+        }
+
+        // Set node before this to this's next node
+        currentNode.Next = this.Next;
+
+        // Clear by setting the node to itself 
+        this.Next = this;
```
BenjaminMichaelis, line 61: I think all you need for this method is this line

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+    public bool Exists(T value)
+    {
+        // create node to navigate with starting with the first node
+        Node<T> currentNode = this;
+
+        // since curNode cannot equal this/first node in the loop, we need to do an 
+        // intial test to make sure the first node does not have the given value.
+
+        // Use of null forgiveness since doc states validation on node values is not nessasary
+        if(currentNode.Value!.Equals(value))
```
BenjaminMichaelis, line 80: What if T is nullable and has a null value? Shouldn't it return true if both Value and value are null?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        // loop until initial node, if values are equal return true, otherwise return false.
+        while(currentNode != this)
+        {
+            // Use of null forgiveness since doc states validation on node values is not nessasary
+            if (currentNode.Value!.Equals(value))
+            {
+                return true;
+            }
+            currentNode = currentNode.Next;
```
BenjaminMichaelis, line 96: A do while loop can reduce this complication

PR #135, opened 2024-02-13
Assignment5 - Alexa Darrington and Johanne McClenahan

JohanneJayde:
Alexa's Github - @alyxadarr
Johanne's Github - @JohanneJayde


Purplepigs0129:
Reviewing

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+    will be able to get back the memory and works with reference objects.
+    After some research as well, it seems the garbage collector works and handles
+    circular references and lists like our case here. 
+    The high level view of how the garbage collector works:
+    Start with locals, statics and GC pinned objects. None of these can be collected
+    Mark every object which can be reached by traversing the children of these objects Collect every object which is not marked.
+    This allows for circular references to be collected just fine. 
+    So long as none of them are reachable from an object known to be uncollectable then the circular reference is essentially irrelevant.
+    */
+    public void Clear()
```
Purplepigs0129, line 60: Great comment on the Clear method

GenericsTests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+    [Fact]
+    public void ToString_NullData_ThrowsInvalidOperationException()
+    {
+        Assert.Throws<InvalidOperationException>(() => new Node<object>(null!).ToString());
+    }
+
+    [Fact]
+    public void ToString_ValidData_ReturnsStringSuccesfully()
+    {
+        Assert.Equal("Alexa", NodeList.ToString());
```
Purplepigs0129, line 26: I'm kind of confused about the whole generic test from what I am seeing it looks like you are but from research, it is saying to have 2 generics in the equal. With how it is written in the assignment it doesn't specify beyond having a generic in the equals. I'm going to mark it as complete for now but I could potentially be wrong but if it is then our tests our wrong too.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    public class Node<T>
+    {
+        public Node(T item)
+        {
+            Data = item;
+            Next = this;
+        }
+
+        public Node<T> Next { get; private set; }
```
Purplepigs0129, line 12: Definitely missed the private set on ours great job!

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+
```
Purplepigs0129, line 5: Same thing here could remove the whitespace

Directory.Build.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <!-- See https://aka.ms/dotnet/msbuild/customize for more details on customizing your build -->
+  <PropertyGroup>
+	  <Deterministic>true</Deterministic>
+	  <LangVersion>12.0</LangVersion>
+	  <ImplicitUsings>enable</ImplicitUsings>
+	  <Nullable>enable</Nullable>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
```
Purplepigs0129, line 10: Whitespace here as well unless its on purpose

GenericsTests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
+using Xunit;
+
+namespace GenericsHomework.Tests;
+
+public class NodeTests
+{
+
+    public Node<string> NodeList { get; }
+
```
Purplepigs0129, line 9: Nice constructor setup for your tests

GenericsTests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        Assert.Equal(newNode, newNode.Next);
+    }
+
+    [Fact]
+    public void Exists_NodeExistsAtEnd_ReturnsTrue()
+    {
+        Assert.True(NodeList.Exists("best!"));
+    }
+
```
Purplepigs0129, line 66: Good testing on seeing if all nodes still exist besides just start and end

Purplepigs0129 (commented):
Define the Node class
Extra Credit

Fundamentals

Didn't see any EC attempted but great job it looks like you covered everything. The only thing is if your tests are actually generic tests or some small whitespace removal. If you are going to attempt the EC best of luck! :)

GenericsHomework/Node.cs
alyxadarr: thank you!

GenericsHomework/Node.cs
alyxadarr: thank you! although I reviewed yours and you guys did do a private set!

rkdeboer:
Reviewing!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+                throw new InvalidOperationException(nameof(Data)); 
+            else
+            {
+                return Data.ToString()!;
+            }
+        }
+
+        public void Append(T item)
+        {
+            if (Exists(item))
```
rkdeboer, line 27: We also used the exists method in our append! Its a great way to reuse code for sure

GenericsTests/NodeTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        Assert.True(NodeList.Exists("Alexa"));
+    }
+
+    [Fact]
+    public void Exists_NodeDoesNotExists_ReturnsFalse()
+    {
+        Assert.False(NodeList.Exists("Rahat"));
+    }
+
+
```
rkdeboer, line 86: Very extensive testing nice work!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    do
+    {
+        if (currentNode.Data!.Equals(item))
+        {
+            return true;
+        }
+
+        currentNode = currentNode.Next;
+    } while (currentNode != this);
```
rkdeboer, line 78: I like the do while usage here too I did not use that when first writing this!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+            Data = item;
+            Next = this;
+        }
+
+        public Node<T> Next { get; private set; }
+        public T Data { get; set; }
+        
+        public override string ToString()
+        {
+            if (Data==null)
```
rkdeboer, line 17: You might be able to change this to a throwifnull statement with the type of exception if you want to! This works too though!

rkdeboer (commented):
Define the Node class
Extra Credit

Fundamentals

GenericsTests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+
```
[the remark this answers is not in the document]
alyxadarr: will do thank you!

GenericsHomework/GenericsHomework.csproj
JohanneJayde: Yes, we will do that :)

Directory.Build.props
JohanneJayde: I believe that was on purpose but would need to check again.

GenericsTests/NodeTests.cs
JohanneJayde: Thanks! I forgot about it but then I saw I was using the same list so I did that instead

GenericsTests/NodeTests.cs
JohanneJayde: Thanks. These tests were originally not there but then got added and helped fix bugs with checking if an item existed within the middle of node list.

GenericsHomework/Node.cs
JohanneJayde: Yes!

GenericsTests/NodeTests.cs
JohanneJayde: Thank you! I know I've had issues with doing LinkedLists in Java. Now that I know how to do tests, its very easy to check to make sure things work properly.

GenericsHomework/Node.cs
JohanneJayde: I actually didn't as well until I checked another person's code and realized why it made sense to use rather than while loop

GenericsHomework/Node.cs
JohanneJayde: Good catch! I need to go and fix the tests for null values and will do this :)

GenericsTests/NodeTests.cs
JohanneJayde: Hi, Assert.Equals in the method right now is checking for strings. I can see why that is confusing since we are using generics. I will probably switch it to Assert.Equals<string> and disable warning to clarify because we are doing generics.

GenericsTests/VennDiagramTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    public void Intersection_NullFirstCircleName_ThrowsArgumentNullException()
+    {
+        Exception execption = Assert.Throws<ArgumentNullException>(() => VennDiagram.Intersection("Null Wizards", null!, "Wizards"));
+    }
+
+    [Fact]
+    public void Intersection_NullSecondCircleName_ThrowsArgumentNullException()
+    {
+        Exception execption = Assert.Throws<ArgumentNullException>(() => VennDiagram.Intersection("Null Wizards", "Wizards", null!)) ;
+    }
```
JohanneJayde, line 156:
```suggestion
    [Fact]
    public void Intersection_NullName_ThrowsArgumentNullException()
    {

        Assert.Throws<ArgumentNullException>(() => VennDiagram.Intersection(null!, "New Yorkers", "Superheroes"));
    }

    [Fact]
    public void Intersection_NullFirstCircleName_ThrowsArgumentNullException()
    {
        Assert.Throws<ArgumentNullException>(() => VennDiagram.Intersection("Null Wizards", null!, "Wizards"));
    }

    [Fact]
    public void Intersection_NullSecondCircleName_ThrowsArgumentNullException()
    {
        Assert.Throws<ArgumentNullException>(() => VennDiagram.Intersection("Null Wizards", "Wizards", null!)) ;
    }

```
These tests don't do anything I believe. You should either Assert that Exception E is type of ArgumentNullException. This can be done using Assert.isType<ArgumentNullException>(exception). Or you just remove setting them to exception. I'd prefer dong the latter since that's how we did it in class before and on previous assignments.

GenericsHomework/Circle.cs
```diff
@@ -0,0 +1,33 @@
+﻿
+namespace GenericsHomework;
+
+    public class Circle<T>(string Name , List<T>? items = null)
```
JohanneJayde, line 4:
```suggestion
    public class Circle<T>(string Name)
```
By having it so you can pass in a premade list, you are creating a bug because what if you sent in a list with all duplicate values. I see that you addressed this in your VennDiagram class. It should also be addressed here.

GenericsHomework/Circle.cs
```diff
@@ -0,0 +1,33 @@
+﻿
+namespace GenericsHomework;
+
+    public class Circle<T>(string Name , List<T>? items = null)
+    {
+        public string Name { get;} = Name ?? throw new ArgumentNullException(nameof(Name));
+        public List<T> Elements { get; private set; } = items ?? [];
```
JohanneJayde, line 7: A way to fix the issue that I raised above would be to use a HashSet<T>. This will make so you have to only have unique values which would allow you to pass a premade set instead of only being able to set one. You also won't have to check for duplicate values within the parameter list

GenericsHomework/VennDiagram.cs
```diff
@@ -0,0 +1,87 @@
...
+    public Circle<T> Intersection(string name, string firstCircleName, string secondCircleName)
+    {
+        ArgumentNullException.ThrowIfNull(name);
+        ArgumentNullException.ThrowIfNull(firstCircleName);
+        ArgumentNullException.ThrowIfNull(secondCircleName);
+
+        List<string> selecteCircles = [firstCircleName, secondCircleName];
+
+        List<Circle<T>> intersectCircles =
+        Circles.Where(circle => selecteCircles.Contains(circle.Name)).ToList();
```
JohanneJayde, line 62:
```suggestion
        List<Circle<T>> intersectCircles =
        Circles.Where(circle => (circle.Name.Equals( firstCircleName,StringComparison.OrdinalIgnoreCase)) || (circle.Name.Equals(secondCircleName, StringComparison.OrdinalIgnoreCase))).ToList();

```

selectedCircles could be removed and instead use two conditions with or in LINQ statement.

GenericsHomework/VennDiagram.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        //If intersection count was less than 2, that means the names are not valid
```
JohanneJayde, line 64: What if the user enters in the same name for firstCircleName and secondCircleName? It will return 1 Circle. S intersectCircles.Count will be 1. This will make it enter into the if statement where it will enter into the else statement and trigger the ArgumentException. Either handle this by making so they both can't have the same name by throwing an exception. I think it'd be best to instead just return a new Circle with the new name with elements of the one found circle.

JohanneJayde (commented):
After looking back over the VennDiagram and Circle class. I found that there were issues that I did not find until after it was due.

GenericsTests/VennDiagramTests.cs
BenjaminMichaelis: I think asserting that the exception is actually the type you expect is valid. Aka, like for an argumentNullException, check that param passed into the exception is what you expect to be thrown for this

GenericsHomework/VennDiagram.cs
```diff
@@ -0,0 +1,87 @@
...
+        foreach(Circle<T> circle in Circles[0..^1])
+        {
+            circles.Append(CultureInfo.InvariantCulture,$"{circle}, ");
+        }
+
+        circles.Append(CultureInfo.InvariantCulture, $"{Circles[^1]}");
+
+        return circles.ToString();
+    }
+    public Circle<T> Intersection(string name, string firstCircleName, string secondCircleName)
```
BenjaminMichaelis, line 53: I am not totally sure on what this `Intersection` does, it's confusing to me

PR #136, opened 2024-02-13
Assignment 5 - Roberto G & Timothy N

timmytaters:
@timmytaters 
@robertogarcia11


Aramirez61:
I'll review this.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+        public void Clear()
+        {
+            // it is not sufficient to only set Next to itself, we need to unlink the current node from the list
+            // garbage collector built into C# will automaticaly collect the rest of the other nodes 
+            Node<T> current = this;
+            current.Next = this;
+
+        }
+
+        public override string ToString()
```
Aramirez61, line 54: The project requirements said override the toString() that "that writes out the value's ToString() result". My partner and I's implementation was to print the current value types string representation and not the entire list. I am unsure whether the goal of the to string was to print the whole list or just the value of the current node from reading the requirements.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class Node<T>
+    {
+        public T Data { get; set; }
+        public Node<T> Next { get; private set; }
+
+        public Node(T data)
+        {
+            Data = data;
+            //Set to this so not nullable
+            Next = this;
```
Aramirez61, line 12: We did this the same way. It ensures Next is always referencing an instance of Node<T>, preventing it from being null. Good stuff.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public void Append(T value)
+        {
+            if (Exists(value))
+            {
+                throw new InvalidOperationException("value already exists");
```
Aramirez61, line 19: Minor nitpick on capitalization of first word in exception message.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            Data = data;
+            //Set to this so not nullable
+            Next = this;
+        }
+
+        public void Append(T value)
+        {
+            if (Exists(value))
+            {
```
Aramirez61, line 18: We threw an ArgumentException here. Not saying yours is wrong, but what is your reasoning for using InvalidOperationException here?

Aramirez61 (commented):
Define the Node class
Add a ToString() override that writes out the value's ToString() result. ❌✔???
(I am unsure if their implementation follows what the requirements are asking for since it didn't mention printing the whole list)
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌✔
(Their Clear method used setting next to itself, so the second requirement does not apply)
Extra Credit

Fundamentals

Overall good work. Your work is well organized your directory build props and targets files. Your implementations were pretty similar to ours, and seeing your to string made me reconsider what we did for our to string, though the project requirements were a little vague on it.

Joshua-Pedersen:
Will review

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,72 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Diagnostics.Metrics;
+using System.IO;
+using Xunit;
```
Joshua-Pedersen, line 5: Un-used Using statements, ``System``, ``System.Diagnostics.Metrics``, ``System.IO``, and ``Xunit`` should be safe to remove.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
+﻿namespace GenericsHomework
```
Joshua-Pedersen, line 1: Not sure how you didn't get a build/test error here, but I believe this should be written as 
``namespace GenericsHomework;`` like you have in your ``NodeTests.cs`` class

GenericsHomework/Node.cs
Joshua-Pedersen: +1 on this, I do believe that it should be returning the string value of a singular node as apposed to the whole list.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+            return false;
+
+        }
+
+        public void Clear()
+        {
+            // it is not sufficient to only set Next to itself, we need to unlink the current node from the list
+            // garbage collector built into C# will automaticaly collect the rest of the other nodes 
+            Node<T> current = this;
+            current.Next = this;
```
Joshua-Pedersen, line 50: Does this not just "set Next to itself"? If there is another node it's ``Next`` is still pointing at the base node.

Joshua-Pedersen (commented):
I don't believe ``.editorconfig`` needs to be pushed

## Instructions

  - Define the `Node` class
  - Add a `ToString()` override that writes out the value's `ToString()` result. ❌(ToString should return a singular node's value.)
    - Whether it is sufficient to only set Next to itself ❌ (Your comments states this is not sufficient but your implementation does so.)
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌✔(Unsure with this implementation)

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals

- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  

GenericsHomework/Node.cs
timmytaters: Understood, I changed it to this.toString()

GenericsHomework/Node.cs
timmytaters: Thank you!

GenericsHomework/Node.cs
timmytaters: Eh nitpicky is fine if its a quick fix lol

GenericsHomework/Node.cs
timmytaters: You make a good point, will change

GenericsHomework.Tests/NodeTests.cs
timmytaters: Fair point, will do so

GenericsHomework/Node.cs
timmytaters: Oh yeah you're right, normally the build catches that

GenericsHomework/Node.cs
robertogarcia11: fixed, thanks for looking out!

GenericsHomework/Node.cs
StrangeRanger: It doens't throw an exception/error because it's a block namespace declaration. Both of the following are valid:

```csharp
namespace GenericsHomework;
```

```csharp
namespace GenericsHomework
{
    // Declarations of classes, structs, interfaces, enums, etc.
}
```

If you are interested, you can find more information about it here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/namespace

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,62 @@
...
+    public void Append(T value)
+    {
+        if (Exists(value))
+        {
+            throw new ArgumentException("Value already exists");
+        }
+
+        Node<T> newNode = new(value);
+        newNode.Next = this.Next;
+        this.Next = newNode;
```
BenjaminMichaelis, line 24:
```suggestion
        newNode.Next = Next;
        Next = newNode;
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,62 @@
...
+        // It is reasonable to set Next to itself since the garbage collector is designed to handle reference objects effectively.
+        // By doing this, we allow the garbage collector to reclaim memory appropriately
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+        if(this == null || this.Data == null || Data.ToString() == null)
+        {
+            return "Null";
```
BenjaminMichaelis, line 56: This seems bad to return, similar to returning empty string.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+#pragma warning disable CS8603 // Possible null reference return. Can suppress since null checked abov
+        return Data.ToString();
+#pragma warning restore CS8603 // Possible null reference return.
+    }
```
BenjaminMichaelis, line 61:
```suggestion
    public override string? ToString()
    {
        if (Value == null) return null;
        return Value.ToString();
    }
```
I think this is better

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,79 @@
...
+        Node<int> newNode = new(1);
+        newNode.Append(2);
+        Assert.AreEqual<Boolean>(true, newNode.Exists(2));
+    }
+    [TestMethod]
+    public void ExistsDoesntExistFalse()
+    {
+        Node<int> newNode = new(1);
+        newNode.Append(3);
+        Assert.AreEqual<Boolean>(false, newNode.Exists(2));
```
BenjaminMichaelis, line 45:
```suggestion
        Assert.AreEqual<bool>(false, newNode.Exists(2));
```
prefer bool over Boolean

PR #137, opened 2024-02-13
Assignment5 - Joshua Merrell and Trevor Felchlin

Tfelchlin13:
github accounts: @Tfelchlin13 and @joshmerrell


EthanAGi:
I can Review!

Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+  <PropertyGroup>
```
Joshua-Pedersen, line 2: You are missing the ``<LangVersion>`` here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,137 @@
...
+                return true;
+            }
+            current = current.Next;
+        }
+        return false;
+    }
+
+    public void Add(T item) => Append(item);
+
+    public bool Contains(T item) => Exists(item);
```
Joshua-Pedersen, line 75: Is there a reason you have these methods? They appear to only call another method.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System.Collections;
+
+namespace GenericsHomework;
+public class Node<T> : IEnumerator<T>, ICollection<T> where T : notnull
```
EthanAGi, line 4: It would seem that this error is thrown because this Object uses ICollection, but does not have a name that suggests that this is a Collextion, (NodeCollection, etc).

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,137 @@
...
+        Value = value ?? throw new ArgumentException(nameof(Value));
+        Next = this; // Points back to itself by default
+        _current = this;
+    }
+    public override string ToString()
+    {
+        return Value.ToString() ?? "";
+    }
+
+    public void Clear()
```
Joshua-Pedersen, line 41: Don't forget to give a comment on ``"...whether you need to worry about garbage collection..."``

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        Next = this;
+    }
+
+    public void Append(T value)
+    {
+        if (Exists(value))
+        {
+            throw new ArgumentException("Duplicate value detected.");
```
EthanAGi, line 50: After your error message, inserting "nameof(value)" might be required.

GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        Node<int> firstNode = new(4);
+        firstNode.Append(5);
+        int[] ints = { 1, 2, 3 };
+        Assert.Throws<ArgumentOutOfRangeException>(() => firstNode.CopyTo(ints, -1));
+        Assert.Throws<ArgumentOutOfRangeException>(() => firstNode.CopyTo(ints, 2));
+    }
+
+    [Fact]
+    public void CopyTo_CopiesValues()
```
Joshua-Pedersen, line 99: This test is failing. Unsure what ``CopyTo`` is intended to do.

EthanAGi (commented):
  - Define the `Node` class
    - Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌ <- No comment

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ✔<- This is implemented, but seems to be implemented incorrectly, and is why the build is crashing.

## Fundamentals

## See [Docs](https://github.com/IntelliTect-Samples/EWU-CSCD371-2024-Winter/tree/main/Docs)

GenericsHomework/Node.cs
Joshua-Pedersen, line 4: A lot of non-needed methods in this class. Multiple of which are not tested.

Joshua-Pedersen (commented):
One test is failing, I assume the extra classes are to implement ``ICollections`` but I think the way in which it was done goes against the ``simplicity over complexity`` goal. Other than that, don't forget to leave the requested comments, update ``<LangVersion>``, and add testing to all of your methods.

## Instructions

  - Define the `Node` class
  - Add a `Clear` method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following:
    - Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ✔

## Fundamentals

- Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional)   ❌
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  
- All of the above should be unit tested ❌
- Choose simplicity over complexity ❌

Directory.Build.props
```diff
@@ -0,0 +1,11 @@
...
+    <TargetFramework>net8.0</TargetFramework>
```
Tfelchlin13, line 3:
```suggestion
    <TargetFramework>net8.0</TargetFramework>
    <LangVersion>12.0</LangVersion>
```

Directory.Build.props
Tfelchlin13: fix thank you

GenericsHomework/Node.cs
Tfelchlin13: oh yea thanks

GenericsHomework/Node.cs
Tfelchlin13: we need them in order to implement the interface ICollection<T>

GenericsHomework/Node.cs
Tfelchlin13: you are right

GenericsHomework/Node.cs
Tfelchlin13: Lot of them are for the interface and are required.

GenericsHomework/NodeCollection.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System.Collections;
+
+namespace GenericsHomework;
+public class NodeCollection<T> : IEnumerator<T>, ICollection<T> where T : notnull
```
BenjaminMichaelis, line 4:
```suggestion
public class NodeCollection<T> : ICollection<T> where T : notnull
```
You should not implement `IEnumerator<T>`. It is not thread safe

GenericsHomework/NodeCollection.cs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public IEnumerator<T> GetEnumerator()
+    {
+        return this;
+    }
+
+    IEnumerator IEnumerable.GetEnumerator()
+    {
+        return this;
```
BenjaminMichaelis, line 111: This won't actually work. A foreach test would prove it

GenericsHomework/NodeCollection.cs
BenjaminMichaelis, line 1: I will give you some EC points for the attempt, but the implementation is not right. `IEnumerator` is giving you the biggest problems

PR #138, opened 2024-02-13
Assignment5 - Spencer West & Matthew Powers

SaaWest:
@SaaWest @Matthew-Powers


joshmerrell:
I will review this

joshmerrell:
Over all it was pretty well done. All of the tests passed, and it has good code coverage.
```
- The `Next` property setter should be private.  ✔
- Place all shared project properties into a `Directory.Build.Props` file. ❌
- All of the above should be unit tested ❌
```

NathanMarsee:
I'll review this one as well.

MannyPoptart (commented):
Instructions

        Define the Node class

Fundamentals

    Place all shared project properties into a Directory.Build.Props file. (This is missing)
    Place all shared project items into a Directory.Build.targets file. (This is missing)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,66 @@
...
+        Node<T> tempNode = new(value);
+        current.Next = tempNode;
+        tempNode.Next = this;
+
+    }
+    public void Clear()
+    {
+        //With the clear method because none of the the disconnected nodes are being referenced we don't have to worry about garbage collection
+        //If Clear() is never called, as long as the list isn't too big then we shouldn't worry about garbage collection
+        //however if the list is too big then we need to handle garbage collection otherwise we will get a stack overflow
```
NathanMarsee, line 49: Good explanation here. We did the clear() method the same way, but your comments are much clearer than ours - nice work!

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+	  <TargetFramework>net8.0</TargetFramework>
+	  <ImplicitUsings>enable</ImplicitUsings>
+	  <Nullable>enable</Nullable>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <LangVersion>12.0</LangVersion>
+	  <IsPackable>false</IsPackable>
+	  <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
```
NathanMarsee, line 11: All this should be included in your Directory.Build.props file, which should be located in the root of your project (with the .sln file and such).
You don't need to remove it from this file, but it should be in the Directory.Build.props file as per the assignment guidelines.

HadiM2:
### **Reviewing : )**

NathanMarsee (approved):
**Checklist:**
- Create a class library project called "GenericsHomework.". ✔
    - Define the Node class ✔
    - The values should be homogeneous. ✔
    - Include a constructor that takes a value. (No validation is necessary on the value). ✔
    - Add a ToString() override that writes out the value's ToString() result. ✔
    - Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
        - The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
        - The Next property setter should be private. ✔
    - Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔

**Extra Credit:** _(not attempted)_

**Fundamentals**
- Place all shared project properties into a Directory.Build.Props file. ❌
- Place all shared project items into a Directory.Build.targets file. _(unnecessary according to Mark)_
- nullable reference types is enabled ✔
- Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
- Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
- For this assignment, always use Assert.AreEqual<T>() (the generic version) ✔


Nice work, you're really just missing the Directory.Build.props file. Get that in there, and you should have everything covered.

GenericsHomework/GenericsHomework.csproj
HadiM2, line 1: ✔

GenericsHomework.Tests/GlobalUsings.cs
HadiM2, line 1: Needed?

GenericsHomework.Tests/GenericsHomework.Tests.csproj
HadiM2, line 1: ✔

GenericsHomework.Tests/NodeTests.cs
HadiM2, line 1: Great branch coverage! ✔

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,66 @@
...
+        Next = this;
```
HadiM2, line 50: I am not sure how the clear is working here? isn't clear supposed to delete every node other than the current node?

GenericsHomework/Node.cs
HadiM2, line 1: Look good, just had a quick question about clear() method.

HadiM2 (commented):
Good shit, make sure to do the directory prop requirement. 
- The `Next` property setter should be private.  ✔
- Place all shared project properties into a `Directory.Build.Props` file. ❌

GenericsHomework/Node.cs
SaaWest: thank you

GenericsHomework/Node.cs
SaaWest: we have a choice
`Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following:

    Whether it is sufficient to only set Next to itself ❌✔
    Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌✔`
We could have the removed items circle back on themselves but because they are no longer referenced the garbage collection kicks in. So in this case we set Next on itself so only the first node is referenced to itself

Directory.Build.Props
BenjaminMichaelis, line 1: This file can be deleted, you have a `Directory.Build.props` already

GenericsHomework.Tests/GenericsHomework.Tests.csproj
BenjaminMichaelis: You should in fact also remove it out of this file unless you are overriding which you are not. It is unnecessary here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+    public bool Exists(T value)
+    {
+        Node<T> current = this;
+        do
+        {
+            if (current.Value!.Equals(value))
```
BenjaminMichaelis, line 58: What if Value is null and value is null?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public override string ToString()
+    {
+        if(Value is null)
+        {
+            throw new ArgumentException(nameof(Value));
+        }
+
+        return Value.ToString()!;
```
BenjaminMichaelis, line 27: Dont throw Exception from `ToString` (https://essentialcsharp.com/overriding-object-members#overriding-tostring)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            throw new ArgumentException("Value already exist "+ (nameof(value)));
+        }
+        if (current.Next != current)
+        { 
+            current = current.Next;
+        }
+        Node<T> tempNode = new(value);
+        current.Next = tempNode;
+        tempNode.Next = this;
```
BenjaminMichaelis, line 43:
```suggestion
        if (Exists(value)) throw new ArgumentException("This item already exists in the list");
        Node<T> newNode = new(value)
        {
            Next = Next
        };
        Next = newNode;
        return newNode;
```
simplified

GenericsHomework.Tests/GenericsHomework.Tests.csproj
SaaWest: I think this is here because of a merge conflict that I didn't resolve properly

GenericsHomework/Node.cs
SaaWest: I see I should have thrown an exception since the value T should be T?.

GenericsHomework/Node.cs
SaaWest: sorry

PR #139, opened 2024-02-13
Assignment5- Matthew adams and Rachael DeBoer

MattA41:
@MattA41 and @rkdeboer


SaaWest:
I will review this

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        //Since there is nothing to reference them, we do not need to worry about what happens to them. They are taken care of in garbage collection
+    }
+    public bool Exists(T value)
+    {
+        ArgumentNullException.ThrowIfNull(value);
+        Node<T> dummy = new(value)
+        {
+            Next = this
+        };
+        while (dummy.Next.Next != this)
```
SaaWest, line 97: I believe that this is skipping nodes and not checking all the nodes

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            if (dummy.Next.Value != null)
+            {
+                if (dummy.Next.Value.Equals(value))
+                {
+                    return true;
+                }
+            }
+            dummy.Next = dummy.Next.Next;
```
SaaWest, line 106: i don't think this is necessary

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        }
+        if (dummy.Next.Value != null)
```
SaaWest, line 109: Why is this here if the above code is checking the same conditions

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        Node<T> temp = current.Next;
+
+        while (temp != current)
+        {
+            Node<T> nextTemp = temp.Next;
+            temp.Next = this; // Disconnect the node
+            temp = nextTemp;
+        }
+
+        current.Next = current;
```
SaaWest, line 85: I think this is all you need to have it clear the rest of the nodes I don't believe the others are necessary.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        return ret;
+    }
+    public void Append(T value)
+    {
+        ArgumentNullException.ThrowIfNull(value);
+        if (Exists(value))
+        {
+            return;
+        }
+        Node<T> dummy = new(value)
```
SaaWest, line 59: I'm a little confused by lines 59 to 66 why set Next to this instance of the object and why use dummy.Next.Next != this instead of dummy.Next != this

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            Next = this
+        };
+        while (dummy.Next.Next != this)
+        {
+            dummy.Next = dummy.Next.Next;
+        }
+
+        Node<T> newNode = new(value);
```
SaaWest, line 68: good job here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+    public T Value { get; }
+    public Node<T> Next { get; private set; }
+
+    public Node(T value)
+    {
+        Value = value;
+        Next = this;
+    }
+
+    public void SetNext(Node<T> next)
```
SaaWest, line 19: this can be commented out or delete since its never used

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+
+        Node<T> current = this;
+        Node<T> temp = current.Next;
+
+        while (temp != current)
+        {
+            Node<T> nextTemp = temp.Next;
+            temp.Next = this; // Disconnect the node
+            temp = nextTemp;
```
SaaWest, line 82: I think this should be nextTemp.Next = temp that way they circle back on each other

SaaWest (commented):
Instructions

        Define the Node class
            Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌

Extra Credit

Fundamentals

    Choose simplicity over complexity ❌

GenericsHomework/Node.cs
rkdeboer: I believe that instead of having a dummy node that just points to this, our dummy points to this in its next variable. This means that when we are checking .Next.Next it will be checking the next node since we start from .Next if that makes sense

GenericsHomework/Node.cs
rkdeboer: Care to elaborate? This is what pushes the next node further I dont know what you mean

GenericsHomework/Node.cs
rkdeboer: I believe my partner @MattA41  can provide more insight on that but I believe it was because he wanted to ensure that we dont mess too much with the order of the list but have the dummy node using .next so the values are not changing.

GenericsHomework/Node.cs
rkdeboer: Thank you!

GenericsHomework/Node.cs
rkdeboer: We were having some issues with warnings about Value being null. This may not be relevant code but I can see if it needs changing.

rkdeboer: Just checked it. Not having that check results in an error about the possibility of value being null so its needed to ensure that a null value is not being added

GenericsHomework/Node.cs
BenjaminMichaelis: You have access to Next though? why do you need this?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+            dummy.Next = dummy.Next.Next;
+
+        }
+        if (dummy.Next.Value != null)
+        {
+
+            ret += dummy.Next.Value.ToString();
+        }
+
+        return ret;
```
BenjaminMichaelis, line 50:
```suggestion
        if (Value == null) return null;
        return Value.ToString();
```
much simpler way

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+    }
+    public void Append(T value)
+    {
+        ArgumentNullException.ThrowIfNull(value);
```
BenjaminMichaelis, line 54: Why can't T be null?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+        if (Exists(value))
+        {
+            return;
```
BenjaminMichaelis, line 57: This doesn't allow the caller to know if something has been appended or not.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+            Next = this
+        };
+        while (dummy.Next.Next != this)
+        {
+            dummy.Next = dummy.Next.Next;
+        }
+
+        Node<T> newNode = new(value);
+        dummy.Next.Next = newNode;
+        newNode.Next = this;
```
BenjaminMichaelis, line 70:
```suggestion
     if (Exists(value)) throw new ArgumentException("This item already exists in the list");
     Node<T> newNode = new(value)
     {
         Next = Next
     };
     Next = newNode;
     return newNode;
```
My suggested implementation

GenericsHomework/Node.cs
BenjaminMichaelis: Even just `Next = this;` in that approach

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        }
+        if (dummy.Next.Value != null)
+        {
+            if (dummy.Next.Value.Equals(value))
+            {
+                return true;
+            }
+        }
+        return false;
```
BenjaminMichaelis, line 116:
```suggestion
        var node = this;
        do
        {
            if (EqualityComparer<T>.Default.Equals(node.Value, value)) return true;
            node = node.Next;
        } while (node != this);
        return false;
```
This seems over complicated and error prone. And what about if value or Value is null? Why not something like this?

PR #140, opened 2024-02-13
Assignment5-Andree-Paul

Aramirez61:
@aramirez61 @pja6 Andree Ramirez and Paul Aguilar


connerhanson:
I'll review this.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+namespace GenericsHomework;
+public class Node<T>
```
connerhanson, line 2: Great job creating the node class, I like your implementation of size into it as well!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        Size++;
+    }
+
+    // Method to set the next node
+    public void SetNext(Node<T> next)
+    {
+        Next = next;
+    }
+
+    public override string ToString()
```
connerhanson, line 28: Also great way to implement the toString, we used string interpolation in ours, but your way seems to work just as well.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+namespace GenericsHomework;
+public class Node<T>
+{
+    // Value of the node
+    public T Value { get; set; }
+
+    // Pointer to the next node
+    public Node<T> Next { get; private set; }
```
connerhanson, line 8: Good job making it non-nullable!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+                Next = this.Next // New node points to the current node's next
+            };
+
+            this.Next = newNode; // Update current node's next to point to the new node
+        }
+    }
+
+    public void Clear()
+    {
+        //I don't think you need to worry about the garbage collector as it should take care of the
```
connerhanson, line 52: Great reasoning, I said the same about the garbage collector collecting unreachable nodes.  I did decide to close the other nodes, but after reading your explanation I don't think it was necessary since they become unreachable either way.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        //unreachable nodes after they were cut off after headNode is reset to itself
+        //for that reason I don't think you need to reset the other nodes to themselves either, as regardless of however many are linked, they are all unreachable
+        //and will be taken care of by the garbage collector
+        Node<T> headNode = this;
+        headNode.Next = this;
+        
+
+    }
+
+    public Boolean Exists(T value)
```
connerhanson, line 62: I like the implementation of a do while loop here!

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
+using System.Xml.Linq;
+
+namespace GenericsHomework.Tests;
```
connerhanson, line 3: All tests look great!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public void Clear()
+    {
+        //I don't think you need to worry about the garbage collector as it should take care of the
+        //unreachable nodes after they were cut off after headNode is reset to itself
+        //for that reason I don't think you need to reset the other nodes to themselves either, as regardless of however many are linked, they are all unreachable
+        //and will be taken care of by the garbage collector
+        Node<T> headNode = this;
+        headNode.Next = this;
```
omar5o9, line 57: Great job on this method, I did the same thing.

omar5o9 (commented):
Define the Node class

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            headNode.Append("second");
+            headNode.Append("third");
+            headNode.Append("fourth");
+            headNode.Clear();
+            Assert.Equal("start", headNode.Next.Value);
+            Assert.False(headNode.Exists("second"));
+            Assert.False(headNode.Exists("third"));
+            Assert.False(headNode.Exists("fourth"));
+
+
```
omar5o9, line 91: Good job on this test

GenericsHomework/Node.cs
Aramirez61: Thank you. Originally we were going to use size to be updated when something is appended and then adjust the size when the clear method was called, however we found it wasn't necessary so we are just going to remove size since we didn't find a use for it.

GenericsHomework/Node.cs
Aramirez61: Thank you! We weren't sure if the specifications wanted us to print the whole list or just a specific values node, but we did the latter implementation and tested that it still works for a single node and multiple nodes.

GenericsHomework/Node.cs
Aramirez61: Thanks! Yeah since there's no way to get to those nodes I believe the garbage collector knows to get rid of them.

GenericsHomework.Tests/NodeTests.cs
Aramirez61: Yes I believe Assert.Equals from x unit is the generic version.

GenericsHomework/Node.cs
Aramirez61: Thanks!

GenericsHomework/Node.cs
Aramirez61: Yup, makes sure the next pointer always refers to itself.

GenericsHomework/Node.cs
pja6: Thanks!

GenericsHomework.Tests/NodeTests.cs
pja6: Thanks!

.DS_Store
BenjaminMichaelis, line 1: This should be ignored in your `.gitignore`

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
BenjaminMichaelis, line 6:
```suggestion

```
Unecessary here since they are in your `.props` file

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+        Value = value;
+       
+        Next = this; // By default, the Next pointer refers back to itself
+     
+      
+    }
+
+    public override string ToString()
+    {
+        return Value!.ToString()!;
```
BenjaminMichaelis, line 22:
```suggestion
    public override string? ToString()
    {
        if (Value == null) return null;
        return Value.ToString()!;

```
Close! What if Value is null?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+            throw new ArgumentException("Value already exists");
+        }
+        else
+        {
+            Node<T> newNode = new(value)
+            {
+                Next = this.Next // New node points to the current node's next
+            };
+
+            this.Next = newNode; // Update current node's next to point to the new node
```
BenjaminMichaelis, line 38: Great job! only simplification
```suggestion
                Next = Next // New node points to the current node's next
            };

            Next = newNode; // Update current node's next to point to the new node

```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public void Clear()
+    {
+        //I don't think you need to worry about the garbage collector as it should take care of the
+        //unreachable nodes after they were cut off after headNode is reset to itself
+        //for that reason I don't think you need to reset the other nodes to themselves either, as regardless of however many are linked, they are all unreachable
+        //and will be taken care of by the garbage collector
+        Node<T> headNode = this;
+        headNode.Next = this;
```
BenjaminMichaelis, line 49:
```suggestion
Next = this;

```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,68 @@
...
+        
+
+    }
+
+    public Boolean Exists(T value)
+    {
+        Node<T> headNode = this; 
+        do
+        {
+            if (headNode.Value!.Equals(value))
```
BenjaminMichaelis, line 59: What if Value and value are both null? Aren't they equal then?

PR #141, opened 2024-02-13
Assignment 5-Hadi and Cynthia

montalcy:
@montalcy @Hadim2


pja6:
Will review

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
...
+{
+    private T _value { get; set; }
+    public Node<T>? next { get; set; }
+    public int Size { get; private set; }
+    private Node<T>? Head { get; set; }
+
+    public Node(T value)
+    {
+        _value = value;
+        if (Size == 1) { next = Head; }
```
pja6, line 15: i think you're implementing a dummy head node but it feels like Head isn't connected to anything here - if you initialize your node and size is zero than next is set to this (the current node) so what is head doing in that situation, is it connected to anything or the current node?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
...
+        else { next = this; };
+    }
```
pja6, line 17: should you do something w/ size here since it's the first time you initialize a node and size is currently 0 - wondering if your if statement should then come after this to check the size and connect the first node to the dummy head in the case where there is only one node in the list?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
+﻿//using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace GenericsHomework;
+
+public class Node<T>
+{
+    private T _value { get; set; }
+    public Node<T>? next { get; set; }
```
pja6, line 8: I thought the ? made next nullable and the property should be non-nullable, does having a nullable field allow the property to be nullable too?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        return "Your linked list: " + list;
+    }
+
+    public void Clear()
+    {
+        if (this.next != null)
+        {
+            this.next = null;
+        }
```
pja6, line 41: if next is non-nullable i don't think you would be able to point it to null 
In our solution we pointed the curNode.Next back to itself using this

-also missing the rationale comment per specifications

GenericsHomework/Node.cs
pja6, line 8: next setter should be private
```suggestion
    public Node<T> next { get; private set; }

```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
...
+        if (Exists(data))
+        {
+            throw new ArgumentException("This already exists in the list");
+        }
+        else
+        {
+            if (Head!.next! == Head)
+            {
+                Head = addition;
+                Size = 1;
```
pja6, line 75: I think this is where you're handling the size logic in the constructor but wouldn't you still need to initialize a node first in the actual program otherwise what would you be calling Append on?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,107 @@
...
+
+
+    }
+
+    public bool Exists(T data)
+    {
+        Node<T>? curr = Head;
+        while (curr != Head)
+        {
+            if (curr!._value!.Equals(data))
```
pja6, line 98: There are a lot of null forgiveness operators, if there's a chance something could be null should there be logic to handle that in the case that it is?

GenericsHomework/Node.cs
montalcy: You're right, our Next should not allow nullability. Fixed that, thank you!

montalcy: Fixed that, thanks for catching that!

GenericsHomework/Node.cs
montalcy: I ended up cleaning that up and fixing this class in general. Was in a crunch for time but I see the bigger picture now. I ended up removing our head.

GenericsHomework/Node.cs
montalcy: I ended up seeing from my test that I needed to initialize to 0. But I ended up using this in Append. I was gonna try and use this later on but I havent gone back to do that so I might get rid of it since I might not even use it.

GenericsHomework/Node.cs
montalcy: Totally right. Should not, I went ahead and just looped it to itself. Totally blanked on the comment, will be adding here soon, thanks!

GenericsHomework/Node.cs
montalcy: Yes, we ended up going back and fixing this method since it wasnt making lots of sense. Now when I initialize another node it increments the size but now starting to question if it has much use in general.

pja6 (commented):
* Create a class library project called "GenericsHomework.". ✔
* Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.
    * Define the Node class
    * The values should be homogeneous.
    * Include a constructor that takes a value. (No validation is necessary on the value). ✔
    * Add a ToString() override that writes out the value's ToString() result. ✔
    * Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
        * The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌
        * The Next property setter should be private. ❌
    * Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔
    * Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following:
        * Whether it is sufficient to only set Next to itself ❌
        * Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
        * Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌
    * Create an Exists method to test to see if a value exists in the list. ✔
    * Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ✔
* You should not rely on any BCL generic classes for your implementation. ✔


* Place all shared project properties into a Directory.Build.Props file.✔
* Place all shared project items into a Directory.Build.targets file.//per Mark in class, not required
* nullable reference types is enabled ✔
* Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
* Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
* and enabled .NET analyzers for both projects ✔
* For this assignment, always use Assert.AreEqual<T>() (the generic version) ✔
* All of the above should be unit tested ✔//more tests were added but they’re causing errors now
* Choose simplicity over complexity ❌✔ //I got a little confused with the Head node stuff to be honest, I’m still not quite sure how or if it’s connecting to everything correctly 

Apologies for the delay in posting this part- this is from the comments I made before other changes were implemented - Overall it looks like you’re basically there - There’s some unique implementation that’s different from how we did our Node class, but your tests should show whether or not that implementation works

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    public void Init_DataHead_Successful()
+    {
+        Node<string> list = new("IAmHead");
+        Assert.Equal("IAmHead", list.GetData());
+    }
+
+    [Fact]
+    public void Init_NullData_Fail()
+    {
+        Node<string> list = new(null);
```
rahat15, line 22: Put a nullable here to get rid of this build error

GenericsHomework.Tests/NodeTests.cs
rahat15, line 1: For the underscores build error, I am not exactly sure why it is there but you could try a lower case first letter

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    private T Value { get; set; }
+    public Node<T> Next { get; private set; }
+    public int Size { get; set; }
+
+    public Node(T value)
+    {
+        Value = value;
+        Next = this;
+        Size = 0;
```
rahat15, line 15: Why do we have a size for node? I didn't quite get the usage of size in the Node class

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return Value;
+    }
+
+    override public string ToString() => Value?.ToString() ?? throw new ArgumentNullException(nameof(Value));
+
+    public void Clear()
+    {
+        Next = this;
+        Size = 1;
```
rahat15, line 28: Is this the size of List or Node?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+        Node<T> addition = new(data);
+
+        if (Exists(data))
+        {
+            throw new ArgumentException($"{data} already exists");
+        }
+
+        addition.Next = this.Next;
+        Next = addition;
+        Size++;
```
rahat15, line 43: Same here, the size is of the List according to me. So I don't think it should be in Node class

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+
+
+    public bool Exists(T data)
```
rahat15, line 48: Good piece of code. Easy to read and follow through

rahat15 (commented):
Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌

GenericsHomework.Tests/NodeTests.cs
HadiM2: Thank you we just needed to remove the underscores from method names.

GenericsHomework/Node.cs
montalcy: Ended up removing it since I didnt implement it later on in any way

GenericsHomework/Node.cs
montalcy: Went and fixed this :)

GenericsHomework/Node.cs
montalcy: We ended up fixing this and were able to take off the null forgiveness :)

GenericsHomework.Tests/NodeTests.cs
montalcy: Thanks!

GenericsHomework/Node.cs
montalcy: I ended up deleting, didnt implement this anywhere useful

HadiM2:
We got the extra credit not related to lab:
Question on Auto generating default class with the public modifier instead of internal, and namespace to be file scoped instead of block scoped.
•	Changing the template for the default class template can be done here
- C:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\ItemTemplates\CSharp\Code\1033\Class

•	public class $safeitemrootname$
- changing the default namespace curly brace behavior can also be done there but we could also change it in a .editorconfig file under the CodeStyle section, Namespace declarations: change from Block Scoped to File Scoped, and viola done.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        Node<T> addition = new(data);
+
+        if (Exists(data))
+        {
+            throw new ArgumentException($"{data} already exists");
+        }
+
+        addition.Next = this.Next;
+        Next = addition;
```
BenjaminMichaelis, line 47:
```suggestion
        if (Exists(data))
        {
            throw new ArgumentException($"{data} already exists");
        }
        Node<T> addition = new(data);
        addition.Next = this.Next;
        Next = addition;

```
minor tweaks: `this` not needed here, and no point in allocating an object if it might not be used and doesn't add value yet

PR #142, opened 2024-02-13
Assignment5 - Rahat Bhatia and Joshua Lester

Joshua-Lester3:
@Joshua-Lester3 and @rahat15


timmytaters:
Starting review

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+    public Node(T element)
+    {
+        Element = element;
+        Next = this;
+    }
+
+
+    public override string ToString()
+    {
+        return Element?.ToString() ?? throw new InvalidOperationException($"{nameof(Element)} is null.");
```
timmytaters, line 17: Good checking if its null with the tertiary operator

timmytaters, line 17: Good checking if its null with the tertiary operator

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+        };
+        Next = node;
+    }
+
+    // You do not need to worry about garbage collection because
+    // there are no static or local variable references that you can
+    // control from this class. It would be on whoever has those references
+    public void Clear()
+    {
+        Next = this;
```
timmytaters, line 39: While this clarifies (correctly) that garbage collection handles the removed items, it doesn't state whether or not the removed items should loop on themselves

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFramework>net8.0</TargetFramework>
```
timmytaters, line 4: Since the Directory.Build.props has all of the shared properties, I believe it isn't necessary for them to be in the .proj file. Doesn't hurt though

GenericsHomework.Tests/VennTests.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Xunit;
+
+namespace GenericsHomework.Tests;
+
+public class VennTests
+{
+    #region Circle Tests
```
timmytaters, line 7: I would separate the circle and venn tests into two different test files

timmytaters (commented):
Overall very good, just a couple specific things for the rubric

Instructions
Define the Node class
Include a constructor that takes a value. (No validation is necessary on the value). 
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔
Fundamentals
Place all shared project items into a Directory.Build.targets file.✔-No targets file, but to my knowledge there aren't any shared items
For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌ - Used Assert.Equal() (I'm not gonna comment on every use of Assert.Equal)

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,86 @@
+﻿using Xunit;
+
+namespace GenericsHomework.Tests;
+public class NodeTests
+{
+    #region Node Tests
```
montalcy, line 6: I've never seen this #region, what is this used for?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,56 @@
...
+public class Node<T>
+{
+    public Node<T> Next { get; private set; }
+
+    public T Element { get; }
+
+    public Node(T element)
+    {
+        Element = element;
+        Next = this;
```
montalcy, line 11: efficient way to use "this"

GenericsHomework/Node.cs
montalcy, line 17: good use of simplicity over complexity

montalcy (commented):
Define the Node class
Include a constructor that takes a value. (No validation is necessary on the value).
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔
Fundamentals
For this assignment, always use Assert.AreEqual() (the generic version) ❌ 

Overall really good job at doing everything neatly and simple. I only had a few questions and the only major thing I saw was the testing Assert.Equals.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace GenericsHomework.Tests;
+public class NodeTests
+{
+    #region Node Tests
+    [Fact]
+    public void Element_NodeInitialized_Success()
+    {
+        Node<string> node = new("Jimmy");
+        Assert.Equal("Jimmy", node.Element);
```
montalcy, line 11: Assert,AreEqual was supposed to be used I believe

GenericsHomework.Tests/NodeTests.cs
rahat15: Just to name the region for the particular name. Defining the block of code that what it does. We can collapse the region while going through longer codes.

GenericsHomework/GenericsHomework.csproj
Joshua-Lester3: True! Just fixed it :)

GenericsHomework.Tests/VennTests.cs
Joshua-Lester3: That makes sense. Fixed.

GenericsHomework/Node.cs
Joshua-Lester3: Thank you!

GenericsHomework/Node.cs
Joshua-Lester3: Thank you! :)

GenericsHomework.Tests/NodeTests.cs
Joshua-Lester3: Assert.Equal is the equivalent of Assert.AreEqual in Xunit. The generic version is also not required.

GenericsHomework/Node.cs
Joshua-Lester3: Got it. I fixed it :)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        Next = this;
+    }
+
+    public bool Exists(T element)
+    {
+        Node<T> temporaryNode = this;
+        do
+        {
+            bool isFound = temporaryNode.Element?.Equals(element) ?? throw new InvalidOperationException(nameof(Element));
```
BenjaminMichaelis, line 50: What if Element is null and so is element? Shouldn't these be equal and therefore return true?

PR #173, opened 2024-03-14
Assignment9+10-Multithreading - Cole Stewart and Matthew Powers

colestewart12:
@colestewart12 @Matthew-Powers


EthanAGi:
Reviewing!

EthanAGi (commented):
This looks super clean! Your guys code looks pretty similar to ours, and we're getting the same results! I looked through with the README and you guys look pretty good! I mentioned the difference in signature later on, but I'm unsure of the difference will cause a problem.

## Instructions

1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ❌✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ❌✔
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌ <- I'm marking it this because your signature doesn't match. I believe your method works, but I'm unsure if this is okay.
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
   NOTE: This method does **NOT** use `Task.Run`. ✔

## Fundamentals

- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔ <- I assume you considered

Assignment.Tests/StringExtensions.cs
```diff
@@ -1,35 +1,38 @@
...
+/// </summary>
+public static class StringExtensions
+{
+    /// <summary>
+    /// Returns true if the string matches the given pattern (case-insensitive).
+    /// </summary>
+    /// <param name="s">The string to match</param>
+    /// <param name="pattern">The pattern to match it against.</param>
+    /// <returns></returns>
+#pragma warning disable CA1711 // Bot is being dumb about the "Ex" part of the method name
```
timmytaters, line 17: Would "Regex" work? That's how I often see it written

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,197 @@
...
+        catch
+        {
+            TaskCanceledException t = new();
+            AggregateException a = new(innerExceptions: t);
+            throw a;
+        }
+    }
+
+    // Problem/task 4
+    public async Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
```
timmytaters, line 59: I believe it's supposed to be an IEnumerable type, which an array is, but doing IEnumerable is more generic

timmytaters (commented):
Overall not much to fix, like Ethan said, but #4 might need some tweaking

Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌ - Did array, not IEnumerable
NOTE:
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.✔
Extra Credit
Fundamentals
Place all shared project items into a Directory.Build.targets file.✔- I think the targets is unnecessary, but no harm

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,197 @@
...
+    {
+        StringBuilder? stringBuilder = new();
+        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+        {
+            PingResult result = await RunAsync(item);
+            lock(stringBuilder)
+            {
+                stringBuilder.AppendLine(result.StdOutput);
+            }
+            return result.ExitCode;
```
colestewart12, line 69:
```suggestion
            await task.WaitAsync(default(CancellationToken));
            return result.ExitCode;
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,198 @@
...
+        catch
+        {
+            TaskCanceledException t = new();
+            AggregateException a = new(innerExceptions: t);
+            throw a;
+        }
+    }
+
+    // Problem/task 4
+    public async Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
```
colestewart12, line 59:
```suggestion
    public async Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
```

Assignment.Tests/StringExtensions.cs
colestewart12: I will take a look, thanks!

Assignment/PingProcess.cs
colestewart12: Thanks

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,198 @@
...
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            return await Task.Run(() => Run(hostNameOrAddress));
+        }
+        catch
+        {
+            TaskCanceledException t = new();
+            AggregateException a = new(innerExceptions: t);
+            throw a;
+        }
```
BenjaminMichaelis, line 55: There seems to be a lot of complexity here. What if you just put `cancellationToken.ThrowIfCancellationRequested();` inside of the Task.Run ?

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,198 @@
...
+    {
+        StringBuilder? stringBuilder = new();
+        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+        {
+            PingResult result = await RunAsync(item);
+            lock(stringBuilder)
+            {
+                stringBuilder.AppendLine(result.StdOutput);
+            }
+            await task.WaitAsync(default(CancellationToken));
```
BenjaminMichaelis, line 69: I'd expect something like this is needed:

```C#
  // /Pseudo code:
            StartInfo.Arguments = item;
            {
                    StartInfo, updateStdOutput, default, default);
            });
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,198 @@
...
+        return new PingResult(total, stringBuilder?.ToString());
+    }
+
+    public async Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, 
+        Action<string?>? progressError, CancellationToken token)
+    {
+        Task<int> task = Task.Factory.StartNew(() =>
+        {
+            Process process = RunProcessInternal(startInfo, progressOutput, progressError, token);
+            return process.ExitCode;
```
BenjaminMichaelis, line 84: Why is the signature being changed? Why not just return the PingResult?

`return new PingResult(process.ExitCode, stringBuilder?.ToString());`

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,198 @@
...
+        }, token, TaskCreationOptions.LongRunning, TaskScheduler.Current);
+
+        int exitCode = await task;
+        return exitCode;
```
BenjaminMichaelis, line 88:
```suggestion
        return await Task.Factory.StartNew(() =>
        {
            Process process = RunProcessInternal(startInfo, progressOutput, progressError, token);
            return process.ExitCode;
        }, token, TaskCreationOptions.LongRunning, TaskScheduler.Current);
```
simplify

Directory.Build.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+	<PropertyGroup>
+		<TargetFramework>.net8.0</TargetFramework>
```
BenjaminMichaelis, line 3: nit: This is one of the only things best left in the project files

Assignment/Assignment.csproj
```diff
@@ -0,0 +1,5 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
BenjaminMichaelis, line 3: TargetFramework is in the proj files, why need it in the props file?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+        Process process = Process.Start("ping", "localhost");
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(0, exitCode);
```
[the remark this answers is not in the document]
Matthew-Powers: Because we didn't know it existed and it's not mentioned anywhere in the book or in any of the repos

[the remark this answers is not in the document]
BenjaminMichaelis: That's because it's not in the original story. It's in the bottom of this file and was originally used in some of these tests.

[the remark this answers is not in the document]
Matthew-Powers: So, I'm noticing that this test was in the original, and it wasn't modified. It also wasn't requested to be changed in any of the instructions. Why would we modify an already existing test that seems to be doing it's job?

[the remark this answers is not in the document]
BenjaminMichaelis: You are correct (btw points weren't marked on things like this, it's just a note :) ). This test didn't need this update. But on like `RunLongRunningAsync_UsingTpl_Success` I would question why it also didn't use that method. It points to the fact that the return of that method probably should of been PingResult, not just an int

PR #174, opened 2024-03-14
Assignment9+10 multithreading Matthew Adams and Anna Mikhailenko

MattA41:
Team: @Anna-Mclenko and @MattA41


Purplepigs0129:
Reviewing :)

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,199 @@
...
+    public async Task<PingResult> RunAsync(
+        IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    {
+        StringBuilder? stringBuilder = new();
+        int total = 0;
+        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+        {
+            PingResult result = await RunAsync(item, cancellationToken);
+            
+            if (result.StdOutput != null)
```
Purplepigs0129, line 61: Looks great here we used a semaphore in ours to help with the string builder

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,188 @@
...
+Reply from ::1: time<*
+Reply from ::1: time<*
+
+Ping statistics for ::1:
+    Packets: Sent = *, Received = *, Lost = 0 (0% loss),
+Approximate round trip times in milli-seconds:
+    Minimum = *, Maximum = *, Average = *".Trim();
+    private void AssertValidPingOutput(int exitCode, string? stdOutput)
+    {
+        Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
```
Purplepigs0129, line 180: I think this is where your check if failing at but don't know why cause it happens to be on an isFalse which is here but ti was given to us unless you modified this in anyway it should be ok but otherwise everything looks great

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,199 @@
...
+        StringBuilder? stringBuilder = null;
+        void updateStdOutput(string? line) =>
+            (stringBuilder??=new StringBuilder()).AppendLine(line);
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
+        return new PingResult( process.ExitCode, stringBuilder?.ToString());
+    }
+    
+    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
+    {
+        return new Task<PingResult>(() => Run(hostNameOrAddress));
```
Purplepigs0129, line 30: Great job leaving this to be very simple

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,199 @@
...
+        
+    }
+
+    public async Task<PingResult> RunAsync(
+        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        try
```
Purplepigs0129, line 37: We also used a try catch here I just figured it would be the best way to catch the error without having to use a bunch of if statements or whatever other way you could implement this

Purplepigs0129: I split this into 2 different methods don't know what probably just got tripped up on the wording but seeing how you guys did it in 1 method would be much simpler

Directory.Build.props
```diff
@@ -1,9 +1,11 @@
...
-    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
-  </PropertyGroup>
+<Project>
+<PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	<NetAnalyzers>latest</NetAnalyzers>
```
Purplepigs0129, line 8: Great Catch here I think I missed these in mine oops

Purplepigs0129 (commented):
First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ❌
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ❌✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Extra Credit
Fundamentals

I don't think I saw any errors in the code but I pointed out with a comment where I think the issue is with your check so that you can either fix it or reset it to get the green check. Everything else looks great code is nice, simple, and easy to read. I didn't see an extra credit attempt but otherwise Great Job!

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,199 @@
...
+            AggregateException ae = new(innerExceptions: e);
+            throw ae;
+        }
+
+    }
+
+    public async Task<PingResult> RunAsync(
+        IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    {
+        StringBuilder? stringBuilder = new();
```
Purplepigs0129, line 55: I also used nullable here but good to question whether or not it is absolutely needed here. Might as well ask the question should this ever be null? Just something to ponder here as I am pondering it in mine

SergeiAshikhmin:
I will review this :)

SergeiAshikhmin (commented):
The code looks good and simple. Great job 👍 

Instructions

Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔
 Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔
 Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ 
with a TaskCanceledException inner exception ✔
 using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔
and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔
 NOTE:
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ✔

Extra Credit

Fundamentals

Assignment/PingProcess.cs
Anna-Mclenko: Thank you!

Assignment.Tests/PingProcessTests.cs
Anna-Mclenko: Yes you are right. It is because Unix will throw the error into stdError instead of stdOutput. after adding stdError into Ping result that fixed it for us

Assignment/PingProcess.cs
MattA41: thank you we spent a while trying to understand what mark wanted

Assignment/PingProcess.cs
Anna-Mclenko: Thank you :)

Directory.Build.props
Anna-Mclenko: Thank you!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,188 @@
...
+         AssertValidPingOutput(output.Result);
+        // Do NOT use async/await in this test.
+        // Test Sut.RunTaskAsync("localhost");
+    }
+
+    [TestMethod]
+    public void RunAsync_UsingTaskReturn_Success()
+    {
+        // Do NOT use async/await in this test.
+        Task<PingResult> task = Task.Run<PingResult>(() => Sut.RunAsync("localhost"));
```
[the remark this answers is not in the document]
Anna-Mclenko: Thank you!

Assignment/PingProcess.cs
MattA41: honestly, you make a good point it would not be null because it is making a new StringBuilder every time, we did it in case the ping returns null for some reason

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,188 @@
...
+        PingResult result = Sut.Run("localhost");
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod]
+    public void RunTaskAsync_Success()
+    //testing to see if the task ended correctly
+    {
+         Task<PingResult> output = Sut.RunTaskAsync("localhost");
+        output.Start();
```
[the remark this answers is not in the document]
MattA41: We used start because we are not running the task in runTaskAsync we are creating and returning the task. there was no way I could find to call run on the returned task to get it to start.

Assignment/PingProcess.cs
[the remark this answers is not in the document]
MattA41: We interpreted what mark wanted as retuning a task and not running it. If we wanted the task to run that is what runAsync was for

Directory.Build.props
```diff
@@ -1,9 +1,11 @@
...
+	<LangVersion>12</LangVersion>
```
BenjaminMichaelis, line 9:
```suggestion
	<LangVersion>12.0</LangVersion>
```

PR #177, opened 2024-03-18
Assignment9+10-Multithreading - Alexa & Paul 

pja6:
@alyxadarr -Alexa
@pja6 - Paul


Aramirez61:
I'll review this

Assignment/PingProcess.cs
```diff
@@ -17,28 +17,36 @@ public class PingProcess
...
             (stringBuilder??=new StringBuilder()).AppendLine(line);
-        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
+        Process process = RunProcessInternal(StartInfo, updateStdOutput,default, default);
         return new PingResult( process.ExitCode, stringBuilder?.ToString());
     }
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+          return new Task<PingResult>(()=> Run(hostNameOrAddress));
```
Aramirez61, line 32: I like how simple this is!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,194 @@
...
+                if (inner is TaskCanceledException)
+                {
+                    throw inner;  //Rethrow TaskCanceledException to satisfy expected exception
+                }
+
+                return false;  //Indicates we have not handled other exceptions, so they are still propagated.
+            });
+
+            // If no TaskCanceledException was found (though one should be), rethrow the original AggregateException.
+            // However, in this structure, we should never hit this point with TaskCanceledException present.
```
Aramirez61, line 134: good comments here

Aramirez61 (commented):
Good job on implementing these methods, a lot of them were simple and easy to follow. Just need to invoke the ThrowIfCancellationRequested() method in the RunAsync method to follow the requirements, and also implement and test RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) to complete all the project requirements.

Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ❌ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌ (Make sure to implement this!) NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌

Extra Credit

Fundamentals

Assignment/PingProcess.cs
```diff
@@ -17,28 +17,36 @@ public class PingProcess
...
     }
+ 
 
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        //  cancellationToken.ThrowIfCancellationRequested();
```
Aramirez61, line 41: make sure to invoke this

Assignment/PingProcess.cs
alyxadarr: thank you!

Assignment/PingProcess.cs
alyxadarr: true, we definitely have some fixing to do! We will implement this, thank you!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+
+[TestClass]
+public class PingProcessTests
+{
+    PingProcess Sut { get; set; } = new();
+    //disabling nullable warnings again for testing purposes
+#pragma warning disable CS8618
+    bool isLinux { get; set; }
+    string pingArgs { get; set; }
+    string OutputExpression { get; set; } = "";
```
BenjaminMichaelis, line 22: This doesn't look like it'd have a warning since it has a default initializer. Now whether it should or not can be argued but...

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+#pragma warning restore CS8618
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        isLinux = Environment.OSVersion.Platform is PlatformID.Unix;
+        (string arg, string exp) = isLinux ? ("-c", OutputExpressionLinux) : ("-n", OutputExpressionWindows);
+        pingArgs = arg;
+        OutputExpression = exp;
+        Sut = new();
```
BenjaminMichaelis, line 32: Interesting approach!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+    }
+    [TestMethod]
+    public void Start_PingProcess_Success()
+    {
+        Process process = Process.Start("ping", $"{pingArgs} 4 localhost");
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
```
BenjaminMichaelis, line 39: Why not use `AssertValidPingOutput`?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        //for github actions 
+        int expectedCode = Environment.GetEnvironmentVariable("GITHUB_ACTIONS") is null ? 0 : 1;
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(expectedCode, exitCode);
```
BenjaminMichaelis, line 48: Why not use `AssertValidPingOutput`?

BenjaminMichaelis, line 48: Not the best approach as you are basically asserting that it will fail on the CI. I do like the Environment Variable check, but I would change what is run in that case to use localhost instead of google.com. Changing an expected exit code is not good.

Assignment/PingProcess.cs
```diff
@@ -3,64 +3,107 @@
...
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        //cancellationToken.ThrowIfCancellationRequested();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress), cancellationToken);
+        return await task;
     }
 
-    async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
```
BenjaminMichaelis, line 53: Why change this signature?

Assignment/PingProcess.cs
```diff
@@ -3,64 +3,107 @@
...
+            stringBuilder.AppendLine(output);
+        }
+        int exitCode = await RunLongRunningAsync(StartInfo, updateOutput, default, cancellationToken);
+        PingResult result = new PingResult(exitCode, stringBuilder?.ToString());
+        return result;
+
+
+    }
+   */
+     public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)
```
BenjaminMichaelis, line 95: Why not return `PingResult`?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
+    {
+        // Use exception.Flatten()
+        CancellationTokenSource cancelSource = new CancellationTokenSource();
+        cancelSource.Cancel();
+
+        Task<PingResult> pingTask = Sut.RunAsync("localhost", cancelSource.Token);
+        try
+        {
+            cancelSource.Cancel();
```
BenjaminMichaelis, line 139: Why is this being canceled twice

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    [TestMethod]
+    async public Task RunLongRunningAsync_UsingTpl_Success()
+    {
+        ProcessStartInfo startInfo = new("ping", $"{pingArgs} 4 localhost");
+
+        int exitCode = await Sut.RunLongRunningAsync(startInfo, null, null, default);
+
+        Assert.AreEqual(0, exitCode);
```
BenjaminMichaelis, line 166: Why not use AssertValidPingOutput?

Assignment/PingProcess.cs
```diff
@@ -3,64 +3,107 @@
...
+    {
+        ProcessStartInfo processStartInfo = UpdateProcessStartInfo(startInfo);
+
+        return Task.Factory.StartNew(() =>
+        {
+            using (Process running = RunProcessInternal(processStartInfo, progressOutput, progressError, token))
+            {
+                running.WaitForExit();
+                return running.ExitCode;
```
BenjaminMichaelis, line 104: Something like:

`return new PingResult(process.ExitCode, stringBuilder?.ToString());`

PR #179, opened 2024-03-19
Assignment9+10 multithreading - Ethan Guerin and Rachel Deboer

EthanAGi:
Rachael's Git - @rkdeboer
Ethan's Git - @EthanAGi


JohanneJayde:
I will review!

Joshua-Pedersen:
will review

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        cancellationTokenSource.Cancel();
+
+        try
+        {
+            task.Wait();
+            Assert.Fail("Expected an AggregateException to be thrown.");
+        }
+        catch (AggregateException ex)
+        {
+            Assert.IsTrue(ex.InnerExceptions.Any(innerEx => innerEx is TaskCanceledException));
```
JohanneJayde, line 125:
```suggestion

            Assert.ThrowsException<AggregateException>(() => task.Wait());

```
This method is checking if an `Aggregate Exception` is thrown. So having it also check the inner exception to make sure its the correct type isn't needed.

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+    {
+
+        PingProcess pingProcess = new();
+        string hostNameOrAddress = "localhost";
+        using CancellationTokenSource cancellationTokenSource = new();
+
+        Task<PingResult> task = PingProcess.RunAsync(hostNameOrAddress, cancellationTokenSource.Token);
+        cancellationTokenSource.Cancel();
+
+        await Assert.ThrowsExceptionAsync<OperationCanceledException>(() => task);
```
JohanneJayde, line 140: This isn't testing if it throws `TaskCancelException`. I do like how you are doing the async version of the throwsException. You should do this method like you do the one above by using try catch to catch the inner exception. I would suggest using `flatten()` instead of any because you are only checking the one exception

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
+        stdOutput = WildcardPattern.NormalizeLineEndings(stdOutput!.Trim());
+        Assert.IsTrue(stdOutput?.IsLike(PingOutputLikeExpression) ?? false,
+            $"Output is unexpected: {stdOutput}");
+        Assert.AreEqual<int>(0, exitCode);
+    }
+    private void AssertValidPingOutput(PingResult result) =>
+        AssertValidPingOutput(result.ExitCode, result.StdOutput);
+
+    private readonly string PingOutputLikeExpression = @"
```
JohanneJayde, line 221: This is for the windows versions. Try running `ping` on codespaces or WSL to see what the output is because that is what is being tested on GitHub Actions

Assignment/Assignment/Directory.Build.props
```diff
@@ -1,9 +1,10 @@
-<Project>
-<PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
-  </PropertyGroup>
+<Project>
```
JohanneJayde, line 1: This file is supposed to be on the root directory of the repo. Because it's not it's not applying to all the tests. Move this to the same level as `Assignments` and `Assignment.Tests`.

Assignment/Assignment.sln
```diff
@@ -1,70 +1,56 @@
...
+		Directory.Build.props = Directory.Build.props
+	EndProjectSection
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Assignment.Tests", "Assignment.Tests\Assignment.Tests.csproj", "{0081689F-904E-4E2F-ADFF-EF85CE769A53}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
```
JohanneJayde, line 19: Delete this. You guys already have one in the root directory. This doesn't do anything.  There is an extra `Assignment` and `Assignment.Tests` within the `Assignments` folder. The only files within this one should be `PingProcess.cs` and `Assignment.csproj`

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    [TestMethod]
+    public void RunLongRunningAsync_Success()
+    {
+        var startInfo = new ProcessStartInfo
+        {
+            FileName = "ping",
+            Arguments = "localhost -n 5",
```
JohanneJayde, line 167: This is for windows. Make sure to handle Linux version as well

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
+        StringBuilder stringBuilder = new();
+
+        try
+        {
+            numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
+        }catch (AggregateException) { }
+        int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+        Assert.AreNotEqual(lineCount, numbers.Count() + 1);
+    }
```
JohanneJayde, line 207:
```suggestion
        try
        {
            numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
                    int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
        Assert.AreNotEqual(lineCount, numbers.Count() + 1);
        }catch (AggregateException) { }

    }
```
This should be here I believe.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,161 +16,96 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
-    private Process RunProcessInternal(
-        Process process,
-        Action<string?>? progressOutput,
-        Action<string?>? progressError,
-        CancellationToken token)
-    {
-        process.EnableRaisingEvents = true;
-        process.OutputDataReceived += OutputHandler;
-        process.ErrorDataReceived += ErrorHandler;
+    private static Process RunProcessInternal(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) {
```
JohanneJayde, line 92: A lot has been deleted from this method. I think you should put a note for why you removed a lot since people maybe can't tell just based on what was removed.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,161 +16,96 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+        process?.WaitForExit();
+        return new PingResult(process?.ExitCode ?? 1, output);
     }
 
-    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
+    public static Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        var tcs = new TaskCompletionSource<PingResult>();
+        Ping pingSender = new();
```
JohanneJayde, line 39: I don't think this is how they wanted us to do this method. We are supposed to Run but asynchronously I believe. The implementation of this method is using a `Ping` class is using built in method to handle things rather than what was provided.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,161 +16,96 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+
+    public static async Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
-    }
+        using Ping ping = new();
+        var reply = await ping.SendPingAsync(hostNameOrAddress);
```
JohanneJayde, line 66: I have the same comment for this method as your `RunAsync` method not being what the intended solution was. I would ask Benjamin or Mark if we are allowed to do this.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,161 +16,96 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
-    }
 
-    async public Task<PingResult> RunLongRunningAsync(
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    public static async Task<PingResult[]> RunAsync(IEnumerable<string> hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        var tasks = hostNameOrAddress.Select(hostName => RunAsync(hostName, cancellationToken));
```
JohanneJayde, line 76: I don't believe this is running in parallel. Having something be done in parallel is different from having it done asynchronously which I think you did in this case.

Assignment/Assignment/PingProcess.cs
```diff
@@ -13,161 +16,96 @@ public record struct PingResult(int ExitCode, string? StdOutput);
...
+        return await Task.WhenAll(tasks);
```
JohanneJayde, line 77: This method doesn't return all the StdOutput's of all the ping requests. It will instead just return the ping status which I think is different.

JohanneJayde (commented):
1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ❌ - I don't think this was done correctly based on assignment specifications. But I could be misunderstanding the requirements.
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ❌ - This is because it is using a `Ping` class rather than what we were given. However this could be wrong if they said it was allowed
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ❌ - your test was calling `RunTaskAsync`, not `RunAsync`
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ❌ with a `TaskCanceledException` inner exception ❌ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ❌and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ❌ respectively.

`RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ` is testing if a `TaskCanceledException ` is thrown instead of an `AggregateException`
`RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` is testing for `OperationCanceledException ` instead of `TaskCanceledException` 
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
   NOTE: This method does **NOT** use `Task.Run`. ✔

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. ❌ - This wasn't done correctly. It should be placed in root directory.
- Ensure nullable reference types is enabled  ❌ - This isn't done because the props file isn't in the correct place so it won't be turned on for `Assignment.Tests`
- Ensure that you turn on code analysis for all projects(EnableNETAnalyzers)  ❌
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ❌
- and enabled .NET analyzers for both projects ❌

All of these X's are due to the `Directory.Builds.props` file not being in the correct spot. If it's not in the root directory, it won't apply to all subdirectories. Nullable, TreatWarningsAsErrors, EnableNETAnalyzers, and CodeAnalysisTreatWarningsAsErrors are all not turned on for `Assignment.Tests` due to this error.

- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔
- All of the above should be unit tested ❌ - Unit testing for some of the method is off. They also are not handling Linux versions.
- Choose simplicity over complexity ❌ `RunAsync` methods are overcomplicated by using built in Ping class instead of using our own method in my opinion

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+    [TestMethod]
+    public void RunAsync_UsingTaskReturn_Success()
+    {
+
+        PingProcess pingProcess = new();
+        string hostNameOrAddress = "localhost";
+
+
+        var task = PingProcess.RunTaskAsync(hostNameOrAddress);
```
Joshua-Pedersen, line 82: Believe this should be calling ```RunAsync``` rather than ```RunTaskAsync```

Assignment/Assignment/Directory.Build.props
Joshua-Pedersen, line 1: This file should be on the same file level as the ```.sln```

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        var res = task.Result;
+
+
+        Assert.IsNotNull(res.StdOutput);
+        Assert.IsTrue(res.ExitCode == 0);
```
Joshua-Pedersen, line 87: Can simplify, no reason to use a ```var``` since your method should be returning a ```Task<PingResult>```. You could even remove the creation of ```res``` and do the following
```suggestion
        AssertValidPingOutput(task.Result);
```

Assignment.Tests/PingProcess.Tests.cs
Joshua-Pedersen, line 140: While not in the directions the comments wanted us to use ```ex.Flatten()```

Joshua-Pedersen (commented):
## Instructions

1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`.  ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ❌✔
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ❌Calls wrong method
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ❌ with a `TaskCanceledException` inner exception ❌ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ❌and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ❌ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
   NOTE: This method does **NOT** use `Task.Run`.

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. Need to place the file on the same level as the ```.sln```
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔
- All of the above should be unit tested ❌ Incorrect calls in tests
- Choose simplicity over complexity ❌ Heavy use of var, affects readability

Assignment/Assignment/Directory.Build.props
EthanAGi: fixed

Assignment.Tests/PingProcess.Tests.cs
EthanAGi: Fixed.

Assignment/Assignment/PingProcess.cs
```diff
@@ -1,8 +1,11 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
```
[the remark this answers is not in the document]
EthanAGi: Fixed

Assignment/Assignment/PingProcess.cs
EthanAGi: I believe I made a change that should fix this.

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        PingResult result = Sut.Run("google.com");
+        Assert.AreEqual<int>(0, result.ExitCode);
+    }
```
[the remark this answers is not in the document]
EthanAGi: Fixed

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+    public void RunTaskAsync_Success()
+    {
+        PingProcess pingProcess = new();
+        string hostNameOrAddress = "localhost";
+
+
+        var pingTask = PingProcess.RunTaskAsync(hostNameOrAddress);
+
+        var result = pingTask.Result;
+        Assert.IsNotNull(result.StdOutput);
```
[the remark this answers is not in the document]
rkdeboer: Thats an idea! I like it, but I think I might keep it the way it is now just to ensure directness in our testing, especially if it could cause issues for Github actions. Thank you though!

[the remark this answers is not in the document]
EthanAGi: Fixed

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        PingProcess pingProcess = new();
+        string hostNameOrAddress = "localhost";
+
+
+        var result = await PingProcess.RunAsync(hostNameOrAddress);
+
+
+        Assert.IsNotNull(result.StdOutput);
+        Assert.IsTrue(result.ExitCode == 0);
```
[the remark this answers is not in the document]
rkdeboer: Added this where it seemed relevant thank you!

Assignment/Assignment/PingProcess.cs
rkdeboer: There was not a lot of clarity in the instructions as to specifically how they wanted this implemented. I rewrote this method thank you!

Assignment/Assignment/PingProcess.cs
rkdeboer: I am not really certain how to fix this. How would you go about it?

Assignment/Assignment/PingProcess.cs
rkdeboer: I tried my best to fix it but I am at a bit of a loss. Any suggestions?

Assignment/Assignment/PingProcess.cs
rkdeboer: I believe a lot of it was removed to ensure RunLongRunningAsync. Can add more detail to this

Assignment.Tests/PingProcess.Tests.cs
rkdeboer: Fixed

Assignment.Tests/Assignment.Tests.csproj
```diff
@@ -1,16 +1,18 @@
...
-	  <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
-  </ItemGroup>
-
-</Project>
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+	<TargetFramework>.net8.0</TargetFramework>
+	<LangVersion>latest</LangVersion>
```
BenjaminMichaelis, line 6:
```suggestion
	<TargetFramework>net8.0</TargetFramework>
```
LangVersion can live in the props file

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        if (Environment.GetEnvironmentVariable("GITHUB_ACTIONS") == "true")
+        {
+            Console.WriteLine("Skipping test in GitHub Actions environment.");
+            return;
+        }
```
BenjaminMichaelis, line 40: Halfway there.... Can just change the run to run localhost instead of google.com

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,216 @@
...
+        PingResult result = Sut.Run("localhost");
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod]
+    public void RunTaskAsync_Success()
+    {
+        string hostNameOrAddress = "localhost";
+        var result = PingProcess.RunTaskAsync(hostNameOrAddress);
+        AssertValidPingOutput(result);
```
BenjaminMichaelis, line 67: not quite....
```suggestion
        Task<PingResult> task = PingProcess.RunTaskAsync("localhost");
        AssertValidPingOutput(task.Result);
```

Assignment.Tests/PingProcess.Tests.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    }
+    [TestMethod]
+    public async Task RunAsync_UsingTaskReturn_Success()
+    {
+        string hostNameOrAddress = "localhost";
+        var result = await PingProcess.RunAsync(hostNameOrAddress);
+        AssertValidPingOutput(result);
```
BenjaminMichaelis, line 75:
```suggestion
        Task<PingResult> task = PingProcess.RunAsync("localhost");
        AssertValidPingOutput(task.Result);
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,185 @@
...
+        
+        StringBuilder? stringBuilder = null;
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
+        return new PingResult(process.ExitCode, stringBuilder?.ToString());
+
+    }
+
+    public static PingResult RunTaskAsync(string hostNameOrAddress)
```
BenjaminMichaelis, line 39: Why are the signatures being changed? It should return the task?

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,185 @@
...
+    {
+        PingProcess pingProcess = new PingProcess();
+        return pingProcess.Run(hostNameOrAddress);
```
BenjaminMichaelis, line 42:
```suggestion
    return RunAsync(hostNameOrAddress);
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,185 @@
...
+    }
+
+    public static async Task<PingResult> RunAsync(
+    string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        PingProcess ping = new();
+        Task<PingResult> task = Task.Run(() => ping.Run(hostNameOrAddress), cancellationToken);
+        return await task;
```
BenjaminMichaelis, line 51:
```suggestion
StartInfo.Arguments = hostNameOrAddress;
        StringBuilder? stringBuilder = null;
        void updateStdOutput(string? line) =>
            (stringBuilder??=new StringBuilder()).AppendLine(line);
        Task<PingResult> task = Task.Run(() =>
        {
            cancellationToken.ThrowIfCancellationRequested();
            Process process = RunProcessInternal(
                StartInfo, updateStdOutput, default, cancellationToken);
            return new PingResult(process.ExitCode, stringBuilder?.ToString());
        });
        return await task.WaitAsync(cancellationToken);
```

PR #181, opened 2024-03-19
Assignment9+10 Manny Aguilera & Paulson Ray Thompson

Purplepigs0129:
Manny's Git - @MannyPoptart


Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+    }
+
+    //4
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    {
+        StringBuilder stringBuilder = new();
+        int total = 0;
+
+        // Semaphore to synchronize access to stringBuilder
+        var semaphore = new SemaphoreSlim(1);
```
SaaWest, line 82: why did you guys decide to go with a semaphore instead of a lock?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            if(aggregateException != null)
+            {
+                throw aggregateException.InnerException!;
+            }
+            throw;
+        }
+    }
+
+    /*[TestMethod]
+    async public Task RunAsync_MultipleHostAddresses_True()
```
SaaWest, line 134: there is a way to get the right result from these test but I'm assuming because of the git problems you commented it out

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+        {
+            try
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                PingResult result = await RunAsync(item, cancellationToken);
+
+                if (result.StdOutput != null)
+                {
+                    // Enter critical section
+                    await semaphore.WaitAsync(cancellationToken);
```
SaaWest, line 94: why did you wait on the cancellation token instead of an int?

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+        } catch
+        {
+            TaskCanceledException taskException = new();
+            AggregateException aggregateException = new(innerExceptions: taskException);
+            throw aggregateException;
+        }
+    }
+
+
+    async public static Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
```
SaaWest, line 58: I'd think I just remove this since its never implemented

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+
+
+    //5
+    public Task<int> RunLongRunningAsync(
+    ProcessStartInfo startInfo,
+    Action<string?>? progressOutput,
+    Action<string?>? progressError,
+    CancellationToken token)
+    {
+        return Task.Factory.StartNew(() =>
```
SaaWest, line 137: i think there is a cleaner way of implementing this by removing the return process.Exitcode and using something like 'return await Task.Factory.StartNew(() => RunProcessInternal(process, progressOutput, ProgressError).Exitcode;'

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder? stringBuilder = null;
+        void updateStdOutput(string? line) =>
+            (stringBuilder??=new StringBuilder()).AppendLine(line);
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
+        return new PingResult( process.ExitCode, stringBuilder?.ToString());
+    }
+
+    //1
+    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
```
SaaWest, line 30: I think for 1,2  and 3 you need a stringBuilder and return a PingResult object

SaaWest (commented):
    Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ❌
        First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ❌
        Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ❌
    Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌ NOTE:
        StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌

Extra Credit

Fundamentals

    Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌✔(target framework is here but notvLangVersion)
    For this assignment, consider using Assert.AreEqual<T>() (the generic version) ✔(only a few have this)
    Choose simplicity over complexity ✔(Expect on number it could be more clean)

Joshua-Pedersen (commented):
## Instructions

1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ❌ Tests are here but method is still default
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
   NOTE: This method does **NOT** use `Task.Run`.

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌

## Fundamentals

- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ❌ Need LangVersion
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔
- All of the above should be unit tested ❌ Tests not working

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,260 @@
...
+    }
+
+
+    async public static Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
+    {
+        StringBuilder? stringBuilder = null;
+        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+        {
+            Task<PingResult> task = null!;
+            // ...
```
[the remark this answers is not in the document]
Purplepigs0129: this one didnt need to be as stated above just going to get rid of it since we don't need this, I didn't actually code this it was given and there was nothing to my knowledge that needed to be changed

Directory.Build.props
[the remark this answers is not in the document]
Purplepigs0129: Thanks will do

Assignment.Tests/PingProcessTests.cs
Purplepigs0129: yep just git addressing today

Assignment/PingProcess.cs
Purplepigs0129: I don't believe we do as the return should handle it properly

Assignment/PingProcess.cs
MannyPoptart: The semaphore was originally placed in here so when multiple tasks occur that they occur one at a time correctly. It was also supposed to help ensure the output was correctly collected without issues.

Assignment/PingProcess.cs
MannyPoptart: The purpose of waiting on the cancellation token instead of an int was to ensure that cancellation requests are properly handled and respected throughout the execution of the async tasks.

Assignment/PingProcess.cs
Purplepigs0129: Great Idea but we find that our solution makes a bit more sense to us in reading capability so we are just going to keep ours.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+
+
+    /*[TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(0, exitCode);
+    }
```
BenjaminMichaelis, line 47: This can be checked with using localhost on the github runner

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+    */
+
+    [TestMethod]
+    public void Run_InvalidAddressOutput_Success()
+    {
+        (string expectedOutput, int expectedExitCode) = IsUnix ? ("ping: badaddress: Temporary failure in name resolution", 2) : ("Ping request could not find host badaddress. Please check the name and try again.", 1);
+        (int exitCode, string? stdOutput, string? stdError) = Sut.Run("badaddress");
+        string actualOutput = IsUnix ? stdError! : stdOutput!;
```
BenjaminMichaelis, line 55: Why not do null checks instead of null-forgiving operators everywhere?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod, TestCategory("RequiresAdmin")]
+    public void RunTaskAsync_Success()
+    {
+        // Do NOT use async/await in this test.
+        // Test Sut.RunTaskAsync("localhost");
+        Task<PingResult> task = Sut.RunTaskAsync("localhost");
+        task.Start();
```
BenjaminMichaelis, line 76: This is not necessary
```suggestion
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        int? lineCount = result.StdOutput?.Split(Environment.NewLine).Length;
+        Assert.AreEqual<int?>(expectedLineCount, lineCount);
+    }
+
+
+
+
+    // Create a test for public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)
+    [TestMethod]
+    public void RunLongRunningAsync_ProcessStartInfo_Success()
```
BenjaminMichaelis, line 189: What is this testing? I don't see RunLongRunningAsync get called

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+
+        }
+    }
+
+
+
+    // Create a test for RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    [TestMethod]
+    public async Task RunAsync_MultipleHostAddresses_True()
```
BenjaminMichaelis, line 175: Not quite, I'm not sure what you changed here

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,233 @@
...
+            return task.Result.ExitCode;
+        });
+
+        await Task.WhenAll(all);
+        int total = all.Aggregate(0, (total, item) => total + item.Result);
+        return new PingResult(total, stringBuilder?.ToString());
+    }*/
+
+    //4
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
```
BenjaminMichaelis, line 81: Think more like this:

```C#
        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
        {
   // /Pseudo code:
            StartInfo.Arguments = item;
            {
                    StartInfo, updateStdOutput, default, default);
            });
            await task.WaitAsync(default(CancellationToken));
            return task.Result.ExitCode;
      });

        await Task.WhenAll(all);
        int total = all.Aggregate(0, (total, item) => total + item.Result);
        return new PingResult(total, stringBuilder?.ToString());
```

Directory.Build.props
```diff
@@ -1,9 +1,9 @@
-<Project>
-<PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
-  </PropertyGroup>
+<Project>
+<PropertyGroup>
```
BenjaminMichaelis, line 2: LangVersion?

PR #182, opened 2024-03-20
Assignment9+10 Multithreading - Nathan Marsee and Johanne McClenahan

JohanneJayde:
Nathan's GitHub: @NathanMarsee
Johanne's GitHub: @JohanneJayde


alyxadarr:
I will review this! :)

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,216 @@
...
+        Process process = new();
+        process.StartInfo.FileName = "ping";
+        process.StartInfo.Arguments = "-c 4 localhost";
+        process.Start();
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+    // For some reason pinging google.com doesn't resolve and no packets are sent back
+    // i don't know if this is a due to GitHub actions because this didn't work on
+    // any other addresses we tried.
```
alyxadarr, line 35: ours wasn't working for google either. We commented ours out as well. After seeing comments on Teams it seems like it was an issue with more peers.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,216 @@
...
+    }
+    */
+
+    [TestMethod]
+    public void Run_LocalHost_Success()
+    {
+        int exitCode = Sut.Run("localhost").ExitCode;
+        Assert.AreEqual<int>(0, exitCode);
+    }
+    //When running this test, stdOutput comes back as Empty
```
alyxadarr, line 51: interesting! We couldn't figure this one out too, so this comments helps out.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,216 @@
...
+        CancellationTokenSource cancelSource = new();
+        cancelSource.Cancel();
+        Task<PingResult> task = Sut.RunAsync("localhost", cancelSource.Token);
+        try
+        {
+            task.Wait();
+        }
+        catch (AggregateException a)
+        {
+            if (a.Flatten().InnerException != null)
```
alyxadarr, line 126: Nice job implementing this check with Flatten and rethrowing the exception if there is some inner exceptions within AggregateException. We used a try catch but it looked a lot more messier than this lol

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,216 @@
...
+        ProcessStartInfo startInfo = new("ping", "-c 4 localhost");
+
+        int exitCode = await Sut.RunLongRunningAsync(startInfo, null, null, default);
+
+        Assert.AreEqual(0, exitCode);
+    }
+
+    // Commented out because it is impossible to PROVE whether code is thread safe or not,
+    // as race conditions and other symptoms of thread-unsafety are undefined behavior.
+    /*[TestMethod]
```
alyxadarr, line 161: Interesting thoughts! I can't say if I agree with this or not, because the topic is a bit confusing to me. I did see this test passing one another peers test but they used a try catch block. I agree that these have trouble passing and the wording is definitely confusing. Wish we had another day in class to ask questions

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,216 @@
...
+    public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
+    {
+        IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
+        System.Text.StringBuilder stringBuilder = new();
+        numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
+        int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+        Assert.AreNotEqual(lineCount, numbers.Count() + 1);
+    }*/
+
+    // Alternate version of the above test, using Assert.ThrowsException, which does not work every time
```
alyxadarr, line 171: Interesting! Good job catching on and commenting them out.

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,186 @@
...
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, 
+        CancellationToken cancellationToken = default)
+    {
+        StringBuilder? stringBuilder = new();
+        int total = 0;
+        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            Task<PingResult> task = Task.Run(() => Run(item));
+            lock (stringBuilder)
```
alyxadarr, line 51: Great job using lock to ensure only one thread executes in that block at a time, We did not use this but I wish we did!

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,186 @@
...
+        Action<string?>? progressError, CancellationToken token)
+    {
+        ProcessStartInfo processStartInfo = UpdateProcessStartInfo(startInfo);
+        return Task.Factory.StartNew(() =>
+        {
+            Process running =
+                RunProcessInternal(processStartInfo, progressOutput, progressError, token);
+            running.WaitForExit();
+            return running.ExitCode;
+        }, token, TaskCreationOptions.LongRunning, TaskScheduler.Current);
```
alyxadarr, line 76: Great job! I sound like a broken record but overall clean efficient code. You can see you guys clearly understood the specifications

alyxadarr (commented):
Overall Amazing Jobs guys! Incredibly clean code that shows you both understood the assignment specifications and handled the linux changes properly!!  
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌ no extra credit (yet)
Fundamentals

Assignment.Tests/PingProcessTests.cs
JohanneJayde: Thanks! I didn't know that this was happening until I tested it on a linux machine. When testing, I thought it was weird that when ping failed on Linux that we would get an empty standard output. Then I tested to see if standard error was logging stuff and saw that was where message was. Looking at Connor and Trevor's PR helped me see that since they had discovered something similar.

Assignment.Tests/PingProcessTests.cs
JohanneJayde: Thank you! I wanted us to have this to show that we tested out the behavior of StringBuilder. Since at the time, we thought that there was a way to right this test out where it would pass always. It turns that there is but that involves uses locks to force synchronization but we decided that was no the point of this test.

Assignment.Tests/PingProcessTests.cs
NathanMarsee: We eventually got it working by changing the expected exit code depending on the environment that it's being run in. It's kinda weird, but it gets the job done.

Assignment.Tests/PingProcessTests.cs
NathanMarsee: Thanks!

Assignment/PingProcess.cs
NathanMarsee: Thanks, we just wanted to make sure the appending didn't get messed up, since it's not thread safe.

Assignment/PingProcess.cs
NathanMarsee: Thanks again!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        // Do NOT use async/await in this test.
+        Task<PingResult> result = Sut.RunTaskAsync("localhost");
+        AssertValidPingOutput(result.Result);
+    }
+
+    [TestMethod]
+    async public Task RunTaskAsync_WithProgress_Success()
+    {
+        string progressAppended = "";
+        var progress = new Progress<string?>((output) => progressAppended += output + Environment.NewLine);
```
JohanneJayde, line 96:
```suggestion
        var progress = new PingProccess<string?>((output) => progressAppended += output + Environment.NewLine);
```
`Progress` should be `PingProgress`.  When I did the last revert, I ended up removing this line. The test passed when it is not consistently passing when we worked on it. The reason why is because the order of progress will may not be ordered when using the regular `Report` method from `Progress` class. This was because of how the `SynchronizationContext` worked with the `Report` method. My understanding is that Report is async so it whichever thread returns its report is not guaranteed to be in a set order due to the Sync Context being related to the thread.

This was remedied by instead deriving `IProgress` in `PingProgress`. Using PingProgress will fix issue with inconsistent testing results.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+
+        int expectedExitCode = Environment.GetEnvironmentVariable("GITHUB_ACTIONS") is not null || IsUnix ? 1 : 0;
+
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(expectedExitCode, exitCode);
```
BenjaminMichaelis, line 49: Not the best approach as you are basically asserting that it will fail on the CI. I do like the Environment Variable check, but I would change what is run in that case to use localhost instead of google.com. Changing an expected exit code is not good.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        (string expectedOutput, int expectedExitCode)  = IsUnix ? ("ping: badaddress: Temporary failure in name resolution", 2) : ("Ping request could not find host badaddress. Please check the name and try again.",1);
+        (int exitCode, string? stdOutput, string? stdError) = Sut.Run("badaddress");
+        string actualOutput = IsUnix ? stdError! : stdOutput!;
+        //In Unix, error is logged to StdError, not StdOutput
+        Assert.IsFalse(string.IsNullOrWhiteSpace(actualOutput));
+        actualOutput = WildcardPattern.NormalizeLineEndings(actualOutput!.Trim());
+        Assert.AreEqual<string?>(expectedOutput, actualOutput, $"Output is unexpected: {stdOutput}");
+        // 2 is the exit code for invalid address in Unix
+        // 1 is the exit code for invalid address in Windows
+        Assert.AreEqual<int>(expectedExitCode, exitCode);
```
BenjaminMichaelis, line 73:
```suggestion
        Assert.NotEqual<int>(0, exitCode);
```
This seems easier if we know 0 is always success? but just an idea to reduce complexity

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        Assert.AreEqual<int>(0, exitCode);
+    }
+    //When running this test, stdOutput comes back as Empty
+    [TestMethod]
+    public void Run_InvalidAddressOutput_Success()
+    {
+
+        (string expectedOutput, int expectedExitCode)  = IsUnix ? ("ping: badaddress: Temporary failure in name resolution", 2) : ("Ping request could not find host badaddress. Please check the name and try again.",1);
+        (int exitCode, string? stdOutput, string? stdError) = Sut.Run("badaddress");
+        string actualOutput = IsUnix ? stdError! : stdOutput!;
```
BenjaminMichaelis, line 66: Why not just do some null checks??

`Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));`

PR #186, opened 2024-03-20
Assignment9+10 multithreading- Omar Facundo and Roberto Garcia

omar5o9:
@omar5o9 and @robertogarcia11


rkdeboer:
Starting my review!

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,234 @@
...
+    /*[TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(0, exitCode);
+    }
+    */
+
+    /*[TestMethod]
+    public void Run_InvalidAddressOutput_Success()
```
rkdeboer, line 42: Are these commented out tests ones you plan to use later or just scrapped?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        CancellationTokenSource cancellationTokenSource = new();
+        cancellationTokenSource.Cancel();
+        Task<PingResult> task = Sut.RunAsync("localhost", cancellationTokenSource.Token);
+        try
+        {
+            task.Wait();
+        }
+        catch (AggregateException aggregateException)
+        {
+            aggregateException = aggregateException.Flatten();
```
rkdeboer, line 120: Good use of .Flatten()

Assignment.Tests/WildCardPattern.cs
```diff
@@ -164,8 +164,8 @@ public WildcardPattern(string pattern, WildcardOptions options)
         /// <returns></returns>
         public static WildcardPattern Get(string pattern, WildcardOptions options)
         {
-            if (pattern == null)
-                throw new ArgumentNullException(nameof(pattern));
+            
+            ArgumentNullException.ThrowIfNull(pattern);
```
rkdeboer, line 168: Nice condensing of the code

Assignment.Tests/WildCardPattern.cs
```diff
@@ -1095,7 +1101,7 @@ private class AsterixElement : PatternElement
             }
         }
 
-        private class MyWildcardPatternParser : WildcardPatternParser
+        private sealed class MyWildcardPatternParser : WildcardPatternParser
```
rkdeboer, line 1104: I am a bit curious why you used sealed?

rkdeboer (commented):
    Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ✔

Anna-Mclenko:
I'lll review as well

Assignment/PingProcess.cs
```diff
@@ -20,47 +21,84 @@ public PingResult Run(string hostNameOrAddress)
...
         Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
-        return new PingResult( process.ExitCode, stringBuilder?.ToString());
+        return new PingResult(process.ExitCode, stringBuilder?.ToString());
     }
 
+    // 1]
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        return Task.Run(() => Run(hostNameOrAddress));
```
Anna-Mclenko, line 32: I like that! simple, concise,  to the point. we did the same thing

Assignment/PingProcess.cs
```diff
@@ -20,47 +21,84 @@ public PingResult Run(string hostNameOrAddress)
...
+
     }
 
-    async public Task<PingResult> RunAsync(
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    // 2]
+    public async Task<PingResult> RunAsync(string hostNameOrAddress)
```
Anna-Mclenko, line 37: why did you get rig of the cancellation token. you were supposed to implement that. -> "Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ❌✔ "

all you needed to do was to create the task with a lambda expression and then add the cancellation token to that...
`Task task = new Task(() => {code;}, cancellationToken);`

Assignment/PingProcess.cs
```diff
@@ -20,47 +21,84 @@ public PingResult Run(string hostNameOrAddress)
...
         {
-            Task<PingResult> task = null!;
-            // ...
+            throw new AggregateException(new TaskCanceledException("Operation canceled."));
+        }
+
+    }
 
+    // 4]
+
```
Anna-Mclenko, line 59: I like this method! we did something similar

Assignment.Tests/WildCardPattern.cs
robertogarcia11: twaaaanks :3

Assignment.Tests/PingProcessTests.cs
robertogarcia11: thanks you!

Assignment/PingProcess.cs
```diff
@@ -20,47 +21,84 @@ public PingResult Run(string hostNameOrAddress)
...
+    public Task<int> RunLongRunningAsync(
+        ProcessStartInfo startInfo,
+        Action<string?>? progressOutput,
+        Action<string?>? progressError,
+        CancellationToken cancellationToken)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        return Task.Run(() =>
```
Anna-Mclenko, line 92: we were suposed to use `Task.Factory.StartNew( lambda expression, cancellation token, TaskCreationOptions.LongRunning, TaskScheduler.Current );` we were supposed to avoid to use Task.Run

Assignment.Tests/PingProcessTests.cs
robertogarcia11: We commented them out just so we could get the check to pass for last night. But, unfortunately, we plan to use it!

Anna-Mclenko (commented):
Instructions

First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ❌ (test commented out)

Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ❌✔

First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ❌ (test commented out)

Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await.❌ (test commented out)

Add support for an optional cancellation token to the PingProcess.RunAsync() signature.  ✔ (done int the RunAsync with array of local hosts param)

Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔ (done int the RunAsync with array of local hosts param)

Test that, when cancelled from the test method, the exception thrown is an AggregateException  ✔ (done int the RunAsync with array of local hosts param)

with a TaskCanceledException inner exception  ✔ (done int the RunAsync with array of local hosts param)

using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔

and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.

StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌ (tried running the tests in codespace and Start_PingProcess_Success method kept failing)

Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ❌ (used Task.Run())

Extra Credit

Fundamentals

Assignment.Tests/WildCardPattern.cs
robertogarcia11: @omar5o9 and I kept getting build errors. To fix this, we had to seal a lot of method headers in this class.

Assignment/PingProcess.cs
robertogarcia11: thank you! ☺️

Assignment/PingProcess.cs
robertogarcia11: thanks !

Assignment/PingProcess.cs
omar5o9: Okay will do thanks

Assignment/PingProcess.cs
omar5o9: Thanks will fix this

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    // baaad test
+    /*[TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        int exitCode = Sut.Run("google.com").ExitCode;
+        Assert.AreEqual<int>(0, exitCode);
+    }
+    */
```
BenjaminMichaelis, line 37: This should be fixed, not commented out. It may be bad right now but can be fixed. Check if you are on github, then use localhost if needed. or just use localhost

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /*
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("-c 4 localhost");
+        AssertValidPingOutput(result);
+
+    } */
```
BenjaminMichaelis, line 62: Why is this commented out?

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,229 @@
...
+        try
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            PingResult result = await Task.Run(() => Run(hostNameOrAddress), cancellationToken);
+            return result;
+        }
+        catch (OperationCanceledException)
+        {
+            throw new AggregateException(new TaskCanceledException());
+        }
```
BenjaminMichaelis, line 53: If the throw happens inside a task, aggregate exception will be thrown, no need to overcomplicate it here

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,229 @@
...
+
+    }
+
+    // Task 4]
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    {
+        int count = 0;
+        StringBuilder stringBuilder = new StringBuilder();
+        var semaphore = new SemaphoreSlim(1);
```
BenjaminMichaelis, line 62:
```suggestion
        var semaphore = new SemaphoreSlim(1,1);

```

While semaphore's aren't necessary here, you should always declare a maximum if you only want 1 thread to enter in at a time.

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            var process = new Process
+            {
+                StartInfo = UpdateProcessStartInfo(startInfo)
+            };
+            RunProcessInternal(process, progressOutput, progressError, cancellationToken);
+            return process.ExitCode;
+        }, cancellationToken);
+
+    }
```
BenjaminMichaelis, line 120: Not quite, something like this:

```C#
public Task<PingResult> RunLongRunningAsync(
    {
        StartInfo.Arguments = hostNameOrAddress;
        return Task.Factory.StartNew(() =>
        {
            cancellationToken.ThrowIfCancellationRequested();
                StartInfo, updateStdOutput, default, cancellationToken);
        }, cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
    }
```

